Issue #4, opened 2020-03-25
Improve metadata retrieval from Metabase


davideanastasia:
I suggest to use /api/database/{id}/metadata instead of /api/database/{id}/fields.

Currently, the code is leveraging the /fields API to retrieve field metadata. However, the response I am getting is like:

`{
"id": 39,
"name": "Companies House No",
"base_type": "type/Text",
"special_type": null,
"table_name": "Dim Account",
"schema": null
}`

with no stable identifier for the field name (name field is basically the display name in Metabase).

I suggest to use the /metadata API instead, which will return a nested json of tables and fields, where a field will look like:

`{
"description": null,
"database_type": "STRING",
"special_type": "type/Company",
"display_name": "Company Number",
"base_type": "type/Text",
}`

(I have summarised the response, which is a lot larger).
I'd be happy to implement this change.

gouline:
Assuming you're talking about renaming a column under Data Model and then not being able to reference it by its stable database name? That makes sense, it's a reasonable change.

Another benefit is that it cuts down on API requests for the initial lookup tables (currently, separate for tables and fields).

PR #6, opened 2020-03-26
Use metadata API endpoint for initial lookup tables

gouline:
Replace `/tables` and `/fields` API calls with one `/metadata` call for initial lookup tables.

Discussion: https://github.com/gouline/dbt-metabase/issues/4

### DIFF
--- a/Makefile
+++ b/Makefile
@@ -16,3 +16,6 @@ clean:
 
 dev-requirements:
 	pip3 install -r dev-requirements.txt
+
+dev-install: build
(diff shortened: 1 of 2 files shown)


gouline:
@davideanastasia I got you started on this one. The lookup tables have been swapped over, just need to test whether posting the changes back still works. 

It currently posts all the fields received from the get with the changes made, I'm considering changing that to posting only the fields that changed. Reason being, `/metadata` endpoint has more fields that might confuse it.

gouline:
Fixed the field posting issue I mentioned above and tested. Have a look and see if that's what you had in mind.

PR #10, opened 2021-02-25
Include hidden tables in Metabase

vitorbaptista:
The tables can be hidden/unhidden anytime, so it's useful to have
the docs always in sync.

Also, we currently fail if any tables in dbt doesn't exist in Metabase.
So, without this change, dbtmetabase doesn't work for instances that
have hidden tables.


gouline (approved):
Makes sense, thanks for your contribution!

PR #11, opened 2021-02-25
Display all model compatibility errors at once

vitorbaptista:
If there are multiple incompatibilities between the dbt docs and
Metabase, displaying them all at once makes it much easier to fix. This
allows the user to fix all errors at once, instead of fixing one,
running `dbt-metabase`, fixing the next one, and so on.

### DIFF
--- a/dbtmetabase/metabase.py
+++ b/dbtmetabase/metabase.py
@@ -95,20 +95,21 @@ class MetabaseClient:
 
         _, field_lookup = self.build_metadata_lookups(database_id, schema)
 
+        are_models_compatible = True
         for model in models:
             model_name = model['name'].upper()
             if model_name not in field_lookup:
                 logging.warn("Model %s not found", model_name)
-                return False
-
-            table_lookup = field_lookup[model_name]
-            for column in model.get('columns', []):
-                column_name = column['name'].upper()
-                if column_name not in table_lookup:
-                    logging.warn("Column %s not found in model %s", column_name, model_name)
-                    return False
+                are_models_compatible = False
+            else:
+                table_lookup = field_lookup[model_name]
+                for column in model.get('columns', []):
+                    column_name = column['name'].upper()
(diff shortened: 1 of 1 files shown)


dbtmetabase/metabase.py
```diff
@@ -95,20 +95,21 @@ def models_compatible(self, database_id: str, schema: str, models: list) -> bool
...
                 logging.warn("Model %s not found", model_name)
-                return False
-
-            table_lookup = field_lookup[model_name]
-            for column in model.get('columns', []):
-                column_name = column['name'].upper()
-                if column_name not in table_lookup:
-                    logging.warn("Column %s not found in model %s", column_name, model_name)
-                    return False
+                are_models_compatible = False
```
gouline, line 103: Once this becomes false, it never goes back to true so just `return False` in both instances and then `return True` at the end. The only change here is the `else` branch (which I'm happy with).

gouline (changes requested):
Small logic change. Thanks for your contribution!

dbtmetabase/metabase.py
vitorbaptista: My intent on this PR was to display all errors in all models and fields, so the user could fix all errors at once instead of fixing one error, running dbt-metabase, fixing another error, running it again, and so on. This helped a *lot* when running dbt-metabase for the first time -- it was a great way to find discrepancies between my dbt docs and the tables in the DB.

If I removed these `are_models_compatible` changes and just kept the `else` branch in this PR, the code's behaviour would be the same as it is now.

gouline: Ah, of course, you're doing it for the logs. Makes sense.

gouline (approved):

vitorbaptista:
Thanks for merging and releasing a new version so quickly (and, of course, the project itself -- it's super useful!)

PR #13, opened 2021-03-10
Add flag to ignore undefined properties

elleFlorio:
Add the `ignore_undefined` flag to ignore properties that are not
defined in the dbt file. If the `ignore_undefined` flag is set
the JSON payload sent to the api request in Metabase will
contain only properties that are not None in the dbt file.

Example:
- ignore_undefined = true and 'special_type' is None:
	Metabase special type won't be changed
- ignore_undefined = false and 'special_type' is None:
	Metabase special type will be removed


elleFlorio:
Sorry, I opened this one just by mistake! Closing it.

PR #15, opened 2021-04-21
Rename `special_type` to `semantic_type`

gouline:
Metabase 0.39.0 renamed `special_type` to `semantic_type`. This PR renames it accordingly:

- dbt meta field `metabase.special_type` is now deprecated, use `metabase.semantic_type` instead (will be removed)
- exporter checks API response for `special_type`/`semantic_type` and applies whichever one is present for backwards compatibility with earlier Metabase versions


pcreux:
Hey! Thank you!

PR #17, opened 2021-05-10
Added support for aliased models and source blocks

z3z1ma:
Added alias support in model parsing. Also added searching for "sources" yml blocks pushing table/column documentation to metabase (defining sources is recommended by dbt best practices and often done in tandem with creating staging models). Allows using "metabase.fk_ref" in sources or models .yml in order explicitly define a metabase relationship to a table which allows us to propagate the FK relationship properly to Metabase regardless of aliasing. metabase.fk_ref is looked for first in meta before falling back to default parser.

Issue #3, opened 2020-03-06
Multiple schemas in project


dubravcik:
We are using multiple schemas specified in dbt_project.yml. It would be great if dbt-metabase follows the schema settings and does not send everything to the schema from the command line.

```
models:
  my_project:
    marketing:
      schema: marketing
    product:
      schema: product
    events:
      schema: events
```

thanks @gouline :)

gouline:
No current plans to implement this as I have no use case for it, but feel free to implement it and contribute. The codebase is pretty small and easy to follow :)

gouline:
@z3z1ma I think your recent PRs implement this, right? Thinking of closing this issue.

z3z1ma:
Yes that's right @gouline . We resolved this for both parsers using a hash key. 

We allow no schema passed so long as user is using manifest parser since we can reliably resolve schema in all cases to construct the schema.table hash. 

Folder parser is multi schema compliant only with multiple invocations (one per schema since we need the arg to construct the hash) and even then I'm sure I could imagine nuances that make manifest parser superior for model propagation.

PR #19, opened 2021-06-13
Parsing of manifest.json as well as better yaml parsing support / alias support

z3z1ma:
Picking up the ball where fernandobrito left off. I have a major enhancement (one of two in the works). Lots of cleaned up code, bug fixes from the other fork, personally tested on my own production replicates and internally at my company, etc. There may be some more changes/commits I add (to readme mostly) but I just wanted to post it to get your eyes and awareness. The other PR will involve auto documenting all exposures revealed in metabase through questions / dashboards and constructing a YML that allows us to sync a DAG with metabase fully including click throughs from dbt docs exposures directly to dashboards.

PR #24, opened 2021-06-25
Fix sync when models are listed under dbt sources

remigabillet:
#17 introduced a bug by collecting models listed under DBT sources.

The problem is that sources' tables are typically not stored under the DBT schema but this script only collected Metabase models under the DB schema specified as an argument.
In consequence, when a "source" model is loaded, the script won't be able to match it with a Metabase model and the script times out.

The solution implemented makes the script "schema-aware" by collecting the schema name explicitly for sources. It also indexes all models from Metabase in order to be able to find models outside the DBT Schema.

The behavior for DBT models outside sources is unchanged. They are assume to be stored under the "schema" passed explicitly as an argument.

This change is a stepping stone towards addressing #3


remigabillet:
@z3z1ma Does this make sense ?

gouline:
I'll leave it to @z3z1ma to review this pull request, since it's based on his change.

However, @remigabillet please fix any mention of dbt to the proper capitalisation "dbt" (not "DBT").

z3z1ma:
Hello @remigabillet 

Let me review this in the evening. Thanks for your contribution here! That last PR was a big push to transition to reading the manifest json artifact which implicitly solves for the schema issue for models and sources alike. 

So because this change is focused on what is now dubbed the folder parser, we just need to make sure these changes sit on top of the codebase in the new PR so we might need to either rebase or re integrate your code on top of the new code after we review it.

gouline:
PR #19 is now merged, please rebase your changes @remigabillet. Once done, @z3z1ma and I can review.

remigabillet:
Sounds good. Great job on wrapping up #19 !

remigabillet:
I'm planning to working on the rebase today.

remigabillet:
I closed this one and replaced it with #27

PR #27, opened 2021-07-01
DbtFolder parser: collect schema info for source tables

remigabillet:
New take on #24 since the underlying code changed so much since #24 was open.

### DIFF
--- a/dbtmetabase/__init__.py
+++ b/dbtmetabase/__init__.py
@@ -105,13 +105,13 @@ def export(
     # Sync and attempt schema alignment prior to execution; if timeout is not explicitly set, proceed regardless of success
     if not metabase_sync_skip:
         if metabase_sync_timeout is not None and not mbc.sync_and_wait(
-            metabase_database, schema, models, metabase_sync_timeout
+            metabase_database, models, metabase_sync_timeout
         ):
             logging.critical("Sync timeout reached, models still not compatible")
             return
 
     # Process Metabase stuff
-    mbc.export_models(metabase_database, schema, models, reader.catch_aliases)
+    mbc.export_models(metabase_database, models, reader.catch_aliases)
 
 
 def main(args: List = None):
--- a/dbtmetabase/metabase.py
+++ b/dbtmetabase/metabase.py
@@ -58,13 +58,12 @@ class MetabaseClient:
         )["id"]
 
     def sync_and_wait(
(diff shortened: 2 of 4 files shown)


remigabillet:
@z3z1ma most of #24 was implemented in #19 so this PR is much smaller.

z3z1ma:
Looking good @remigabillet thanks for rebasing. I will review more deeply this later today with any code comments (its a minor change so it wont be too bad). One question that came to mind right away that we already know, makes the folder parser a less robust option in comparison to manifest parsing, is a block like this (I personally use this in a production config). Let me know thoughts on this.

```
version: 2
sources:
- name: modelling
  schema: '{{ target.schema }}'
  loader: python
  loaded_at_field: model_ran_at
  tables:
  - name: fitness_model_3
    identifier: model_gen_f_daily
    description: '# Modelling data for Gen E -> Gen G Hydropanels using a linear model based on TMY solar data'

```

remigabillet:
@z3z1ma Good question. I feel that handing jinja variable substitutions is beyond the scope for DbtFolder. It's inherently limited. At that point, the manifest file should be used.

z3z1ma (approved):
Looks good,
I think until we deprecate folder parser- something like this
```
if "{{" in source_schema_name  and "}}" in source_schema_name:
  logging.warning("dbt Folder Reader cannot resolve jinja expressions- use the Manifest Reader instead.")
``` 
could be useful for end users

We then have a choice to fail fast there or allow the processing to continue with the jinja source name which of course wont match to any schemas in Metabase and thus wont cause any harm (no risk really). 

Not needed in this PR, just a thought if you wanted to add. The next PR should include a deprecation message for folder reader.

z3z1ma (changes requested):
Ah before we proceed, we need to make sure any call to `build_metadata_lookups` does not pass in schema as an arg. It is being passed in currently in the function calls.

remigabillet:
@z3z1ma thanks for the suggestions!

gouline:
Looks good to me, if @z3z1ma is happy then I'll merge.

z3z1ma:
@gouline 

Yep, looks great. I will rebase my fork after this is implemented.

remigabillet:
Thanks @gouline and @z3z1ma!

gouline (approved):

Issue #22, opened 2021-06-24
Parse exposures from Metabase


fernandobrito:
Moving the discussion to an issue to avoid polluting the discussion on the PR.

@z3z1ma wrote on https://github.com/gouline/dbt-metabase/pull/19 that he has a follow-up PR in mind on adding a feature to automatically create dbt exposures based on Metabase questions, making it possible to see on the dbt lineage graph which Metabase questions are using which dbt models. More details on this message as well: https://github.com/gouline/dbt-metabase/pull/19#pullrequestreview-682391872

Just today I open-sourced an internal project that does the same, but between dbt and Tableau: https://github.com/voi-oss/dbt-exposures-crawler.

One thing I would like to ask is what's @gouline scope for `dbt-metabase`. Do you want to keep it narrow and focus on documentation syncing, or do you want to be a comprehensive Metabase <> dbt toolkit and think that @z3z1ma feature would fit on `dbt-metabase` as well?

@z3z1ma, how much of the existing `dbt-metabase` codebase you actually use in the exposures feature? Mostly the Metabase API client I guess?

gouline:
Narrow scope but not just documentation syncing. What you're describing fits fine.

What I want to avoid:
- "Suites" of mandatory features. If somebody just wants documentation sync, that should be easy without enabling anything else. 
- Hacks people use with dbt/Metabase not intended by either project. They might be useful but should go in separate repos. 

Otherwise, no objections from me.

remigabillet:
Thank you @gouline for open-sourcing this work. I love this project. We're just starting a data team at AngelList Talent and I think `dbt-metabase` is going to be a critical part of making DBT+Metabase successful here.

I 👍  the idea of importing exposures from Metabase to DBT. It will greatly improve the DBT experience.

Since you're talking about scope, I thought I would share an idea I've been working on.

Metabase Metrics (and in particular custom aggregations) are very useful to our end-users. I'm finding myself spending a lot of time in Metabase managing Metrics, the UI is very inconvenient.

I'd like to be able to define a list of Metabase Metrics on DBT models as schema properties. I wonder if this is part of this project's scope. The primary challenge is parsing the Metabase expressions which are defined as string (ex: `Distinct(case([step] = "hired", [candidate_user_id]))` into trees that the API can handle. I think it's easiest to maintain if I compile the JS parser from metabase source code into a small node script that can be committed and called from this project. 

[...]

PR #28, opened 2021-07-03
Exposure Parsing

z3z1ma:
# Description

Parsing exposures from Metabase automatically in relation to a dbt project. This function is huge for understanding exposures within your data model. Close the loop between model composition and tangible BI exposures of those models creating an extremely valuable company asset in the dbt docs. In the nature of this project, the aim is we can use almost the same invocation parameters users set up in CI / CD for `dbt-metabase export` as they can with `dbt-metabase exposures`. This lets you run the invocation in production to essentially keep you docs in full sync with BI.

Note: Just opening this for visibility and preliminary code introspection. A bit more to do (but not too much). 

## Type of change

- [x] New feature (non-breaking change which adds functionality)
- [x] This change requires a documentation update

# How Has This Been Tested?

Open to ideas but ultimately I think making a mock api from metabase running on top of a database that has been seeded and ran through with the dbt jaffle shop project would be most robust. Just a lot of work. 

- [ ] CI test for exposure documentation

**Test Configuration**:
* Python version: 3.8

# Checklist:

- [x] My code follows the style guidelines of this project
- [x] I have performed a self-review of my own code
- [x] I have commented my code, particularly in hard-to-understand areas
- [ ] I have made corresponding changes to the documentation
- [x] My changes generate no new warnings
- [x] I have added tests that prove my fix is effective or that my feature works
- [x] New and existing unit tests pass locally with my changes

References #22


dbtmetabase/__init__.py
```diff
@@ -69,9 +85,6 @@ def export(
         assert (
             schema and not schemas_excludes
         ), "Must target a single schema if using yaml parser, multiple schemas not supported."
-    assert bool(schema) != bool(
```
fernandobrito, line 72: This is actually fixing a bug that is available on existing `master` docs sync, right?

z3z1ma: I actually had a bit of confusion there and Im not fully settled on this particular change. But I'll offer what I know and curious as to your comments.

Schema is a required arg on `export_models` function as well as the `build_metadata_lookups`. In that function, the `schema` arg restricts documentation/foriegn_key/semantic_type propagation outwards to only Metabase tables that matched the passed schema. Author likely intended separate invocations per schema? This is independent of the dbt folder / manifest reader and I believe this was in place prior to our work. So if you ran it on your fork without passing schema (took another look at the fork to be sure), a string definitely needed to be passed to `export` and to `build_metadata_lookups`. 

So my question is it seems we have sort of two different intentions running side by side here. If someone can clarify the best case desired behavior and can correct (in the metabase client would be the place most likely).

z3z1ma:
Working on documentation updates now
Then will move onto some pretty cool unit tests

Code is in a reviewable state as is though

z3z1ma:
image

image


Example Native query documentation parsed from metabase into exposure with verbose documentation auto-generated.


image

Example DAG of exposures on a single focused node.

Issue #30, opened 2021-07-10
Fix Boolean CLI Arguments


erika-e:
See discussion on [PR#29](https://github.com/gouline/dbt-metabase/pull/29).

The current CLI arguments `--mb_https` and `--sync` can only be set to false from the command line by passing an empty string. This isn't intuitive.

Suggested changes:
* Remove `mb_https` and propagate `mb_http` downstream to make programmatic interface consistent with CLI interface
* Apply a similar change to `--sync`

PR #31, opened 2021-07-11
Argument naming, ordering and logic consistency

gouline:
Big refactor of CLI/programmatic entry point arguments, to achieve some readability and consistency.

- Arguments only belonging to dbt or Metabase are clearly prefixed with `dbt_` or `metabase_` respectively (common arguments are not prefixed)
- Prefix `mb_` becomes `metabase_` for clarity
- Optional boolean arguments should be false by default, e.g. if we mandate HTTPS to be the default (as we should because it's 2021), optional override should be `use_http`, instead of `use_https` that's true by default overridable to false
- Starting to use the new default format for header comments in VS Code, not going to replace all existing ones (too lazy, maybe in the future), but any new ones or significantly reworked ones are easier to re-create than edit, e.g. see `export()`

We're making a lot of breaking changes with the next release that will feature manifest parsing, we may as well break anything else we need in the process to go along with it.

(Addresses #30 among other things)

### DIFF
--- a/dbtmetabase/__init__.py
+++ b/dbtmetabase/__init__.py
@@ -21,10 +21,10 @@ def export(
     dbt_path: str = "",
     schema: str = "public",
     schemas_excludes: Iterable = None,
-    mb_https: bool = True,
+    mb_use_http: bool = False,
     mb_verify: Union[str, bool] = True,
-    sync: bool = True,
-    sync_timeout: int = None,
+    mb_sync_skip: bool = False,
+    mb_sync_timeout: int = None,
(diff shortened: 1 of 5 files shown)


gouline:
FYI: @z3z1ma, @fernandobrito, @remigabillet, @erika-e

erika-e:
Should the version number change, since these are breaking changes? Otherwise looks good to me!

gouline:
Version changes are done outside of feature PRs. The next version will include quite a few changes.

PR #32, opened 2021-07-14
Fix for new pagination envelope in Metabase 0.40.0

gouline:
Since Metabase 0.40.0, API responses are encapsulated in:

```
{
    "data": [ "OLD RESPONSE HERE" ],
    "limit": 50,
    "offset": 0,
    "total": 2
}
```

This change checks if "data" key is present in the root and uses that instead. I will revisit this later, if I need to cater for pagination issues in requests.

### DIFF
--- a/dbtmetabase/metabase.py
+++ b/dbtmetabase/metabase.py
@@ -470,4 +470,10 @@ class MetabaseClient:
                 raise
         elif not response.ok:
             return False
-        return json.loads(response.text)
+
+        response_json = json.loads(response.text)
+
+        # Since X.40.0 responses are encapsulated in "data" with pagination parameters
+        if "data" in response_json:
+            return response_json["data"]
+        return response_json


gouline:
This change will be backported and released as 0.7.1 to unblock anyone who upgraded Metabase, but not ready to migrate to 0.8.0 coming out soon with some breaking changes.

gouline:
This change has been backported and released as [0.7.1](https://pypi.org/project/dbt-metabase/0.7.1/) to unblock anyone who upgraded their Metabase, but not ready to migrate to 0.8.0 to be released soon with some breaking changes.

fernandobrito:
Thanks for the good PR description. One of our in-house automation (deactivate inactive Metabase users) started failing after upgrading to 0.40 and your description here made it very easy to understand exactly what has changed 😄 .

PR #55, opened 2021-09-04
Restore Python 3.6+ compatibility

### LINKED ISSUE #52: Minimum Python required version is not correct
Hi!

The docs state that this amazing library "Requires Python 3.6 or above", and also `setup.py` states the same.

However, some recently introduced changes (https://github.com/gouline/dbt-metabase/blob/master/dbtmetabase/models/metabase.py#L28) broke that. Using `typing.Literal` requires Python 3.8+. Running the library with Python 3.7 for example will raise:

```
ImportError: cannot import name 'Literal' from 'typing' (/Library/Frameworks/Python.framework/Versions/3.7/lib/python3.7/typing.py)
```

Alternatives on what to do:
* Remove this `Literal` type hint and instead use Enum
* Install an extra dependency that will make this type hint work in 3.6+ (see info box on https://mypy.readthedocs.io/en/stable/literal_types.html)
* Bump the minimum required version to 3.8. Maybe a bit overkill just because of a type hint. dbt itself doesn't run well on 3.9, so requiring a minimum of 3.8 leaves a very tight compatibility range. 

Orthogonal to that, we could:
* Explicitly set a Python version (3.6) in the GitHub Actions workflow that runs CI tests using:
```
- uses: actions/setup-python@v2
  with:
    python-version: '3.6.x' 
```
* or, Test multiple Python versions with GitHub Actions matrix https://github.com/actions/setup-python#usage
* or, Test multiple Python versions with `tox`

I'm happy to contribute with a PR once the maintainer and contributors align on how to move forward 😄 .

fernandobrito:
Resolves #52

- [x] Reverts using `Literal` so the project runs on Python 3.6+
- [x] Uses Python 3.6.x on the GitHub Actions to make sure the project runs on Python 3.6+

### DIFF
--- a/.github/workflows/master.yml
+++ b/.github/workflows/master.yml
@@ -14,6 +14,10 @@ jobs:
     steps:
       - uses: actions/checkout@v2
 
+      - uses: actions/setup-python@v2
+        with:
+          python-version: '3.6.x'
+
       - name: Requirements
         run: make requirements
 
--- a/.github/workflows/master.yml
+++ b/.github/workflows/master.yml
@@ -18,9 +18,6 @@ jobs:
         with:
           python-version: '3.6.x'
 
-      - name: Build requirements
-        run: pip3 install "setuptools>=45" wheel
-
       - name: Requirements
         run: make requirements
 
--- a/.github/workflows/master.yml
+++ b/.github/workflows/master.yml
@@ -18,6 +18,9 @@ jobs:
         with:
           python-version: '3.6.x'
 
+      - name: Build requirements
+        run: pip3 install "setuptools>=45" wheel
+
       - name: Requirements
         run: make requirements
 
--- a/.github/workflows/pull_request.yml
+++ b/.github/workflows/pull_request.yml
@@ -11,6 +11,10 @@ jobs:
     steps:
       - uses: actions/checkout@v2
 
+      - uses: actions/setup-python@v2
+        with:
+          python-version: '3.6.x'
+
       - uses: actions/cache@v2
         with:
           path: ~/.cache/pip
--- a/.github/workflows/pull_request.yml
+++ b/.github/workflows/pull_request.yml
@@ -22,6 +22,9 @@ jobs:
           restore-keys: |
             ${{ runner.os }}-pip-
 
+      - name: Build requirements
+        run: pip3 install "setuptools>=45" wheel
+
       - name: Requirements
         run: make requirements
 
--- a/.github/workflows/pull_request.yml
+++ b/.github/workflows/pull_request.yml
@@ -22,9 +22,6 @@ jobs:
           restore-keys: |
             ${{ runner.os }}-pip-
 
-      - name: Build requirements
-        run: pip3 install "setuptools>=45" wheel
-
       - name: Requirements
         run: make requirements
 
--- a/dbtmetabase/utils.py
+++ b/dbtmetabase/utils.py
@@ -1,5 +1,4 @@
 import logging
-import importlib.metadata
 
 
 def get_version() -> str:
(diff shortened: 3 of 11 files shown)


fernandobrito:
(oh ops, this PR was supposed to go against my own fork while I test the GitHub Action 😄 . Not ready yet for review and strange that I can't convert o a WIP PR)

fernandobrito:
I need some help with this one. I can't build the project on Python 3.6.

image

It seems the `setuptools` bundled with this version is too old for `setuptools_scm`, and you can't add `setuptools` as a `setup_requires`. One of the suggestions from the error message is to explicitly install the setup tools (like I am doing in this PR in the GitHub Action to test it out), but what are your thoughts on that?

fernandobrito:
Another place where I need feedback: the package version during runtime magic that you use from https://pypi.org/project/setuptools-scm/#retrieving-package-version-at-runtime requires `importlib.metadata` which is only available on Python 3.8+. The page points to a library for backward compatibility. Do we want to use it?

gouline:
I converted this to draft for you, click "Ready for review" when it's no longer WIP.

fernandobrito:
Ready to be reviewed now.

* For the use of `Literal`, I've replaced it with an Enum
* For the issue with `setuptools_scm` from my screenshot above, I took the suggestion from the error message and installed an appropriate version of `setuptools` on GitHub Actions
* For the issue on`importlib` not being available before Python 3.8, I wrapped it on a `try` block, which worked well locally.

But I'm more than open to suggestions on different approaches to solve those 3 issues.

Also, is it fine if I "Squash and merge" once it's time to merge? The commit history in the branch is a bit messy 😄

.github/workflows/master.yml
```diff
@@ -14,6 +14,13 @@ jobs:
     steps:
       - uses: actions/checkout@v2
 
+      - uses: actions/setup-python@v2
+        with:
+          python-version: '3.6.x'
+
+      - name: Build requirements
+        run: pip3 install "setuptools>=45" wheel
```
gouline, line 22: This should be added to test requirements instead.

.github/workflows/pull_request.yml
```diff
@@ -11,13 +11,20 @@ jobs:
...
+
       - uses: actions/cache@v2
         with:
           path: ~/.cache/pip
           key: ${{ runner.os }}-pip-${{ hashFiles('**/requirements.txt') }}
           restore-keys: |
             ${{ runner.os }}-pip-
 
+      - name: Build requirements
+        run: pip3 install "setuptools>=45" wheel
```
gouline, line 26: Ditto.

dbtmetabase/utils.py
```diff
@@ -16,9 +15,19 @@ def get_version() -> str:
     except ModuleNotFoundError:
         logging.debug("No _version.py found")
 
+    # importlib is only available on Python 3.6+
```
gouline, line 18: You can do this instead of catching for an error https://github.com/gouline/dbt-metabase/blob/master/setup.py#L6

gouline (changes requested):
Left a couple of initial comments. 

@z3z1ma should also review, since most of it is his code.

z3z1ma (approved):
Everything looks accounted for on the Enum update throughout the package. Great job & thanks.

.github/workflows/master.yml
fernandobrito: Smart! 👍

dbtmetabase/utils.py
fernandobrito: Isn't "Ask for forgiveness, not permission" a bit more Pythonic? But I agree that the nested try-except here was not looking that great and now it's a bit cleaner 😄 .

gouline (approved):
LGTM, thanks for your efforts 🚀

PR #56, opened 2021-09-04
Set explicit utf-8 encoding when opening files

fernandobrito:
Addressing `pylint` errors:

image

By default, Python uses the local encoding and on Windows it is not UTF-8.


gouline:
Thanks! We need somebody to double check Windows for us, I'm on Mac/Linux and I think so is @z3z1ma.

gouline (approved):

Issue #63, opened 2021-09-16
What's the supported dbt version?


danielsepulvedab:
Hey, thanks for sharing you work on this. 

I'm having some problems using the library to parse my project. Neither the the manifest parser nor the `DbtFolderReader` is able to parse it without failing. For my project I'm using dbt `0.20.0`, is this version supported by this package? The manifest schema has slightly changed the last couple of versions and I'm wondering if you might be working with another version.

z3z1ma:
Hey @danielsepulvedab 

I use dbt version `0.20.x` myself and have no issue. Is there any more specific information you can share on the error you are seeing?

Warm Regards,

danielsepulvedab:
Hey @z3z1ma, when I use the `--dbt_path` option I'm getting: 
```
Traceback (most recent call last):
  File "/Users/daniel.sepulveda/work/dbt-models/venv/bin/dbt-metabase", line 6, in <module>
    dbtmetabase.main(sys.argv[1:])
  File "/Users/daniel.sepulveda/work/dbt-models/venv/lib/python3.7/site-packages/dbtmetabase/__init__.py", line 364, in main
    dbt_include_tags=parsed.dbt_include_tags,
  File "/Users/daniel.sepulveda/work/dbt-models/venv/lib/python3.7/site-packages/dbtmetabase/__init__.py", line 83, in models
    docs_url=dbt_docs_url,
  File "/Users/daniel.sepulveda/work/dbt-models/venv/lib/python3.7/site-packages/dbtmetabase/parsers/dbt_folder.py", line 79, in read_models
    model_key=ModelKey.nodes,
  File "/Users/daniel.sepulveda/work/dbt-models/venv/lib/python3.7/site-packages/dbtmetabase/parsers/dbt_folder.py", line 130, in _read_model
    mb_columns.append(self._read_column(column, schema))
  File "/Users/daniel.sepulveda/work/dbt-models/venv/lib/python3.7/site-packages/dbtmetabase/parsers/dbt_folder.py", line 174, in _read_column
    for test in column.get("tests", []):
TypeError: 'NoneType' object is not iterable
```

When trying to parse the manifest, I'm getting:

```
Traceback (most recent call last):
  File "/Users/daniel.sepulveda/work/dbt-models/venv/bin/dbt-metabase", line 6, in <module>
    dbtmetabase.main(sys.argv[1:])
[...]

PR #64, opened 2021-09-18
Add some resiliency to parsers processsing tests nodes

z3z1ma:
A PR to address Issue #63 

Adds a simple conditional and an exception catch to two parts of the parser codes which can encounter edge cases like those presented in the issue.


gouline (approved):
LGTM 👍

PR #65, opened 2021-09-26
feat: Refactor CLI - Simplify Config - Allow Env Vars - Add Config Command

z3z1ma:
# Description

Large refactor to use click as cli interface, improve and further abstract interface complexity, allow env vars implicitly, improve logging with global logger + rich, add config yaml with interactive setup via config command, load config via hook in CLI, split commands in sub commands, optimize help text, add implicit file / folder existence checking on file / folder args. 

Sub commands broken out:

image

Optimized help text

image

Interactive config file generation:

image

Checking config file contents with simple `dbt-metabase config --inspect` and running a CLI command with just `dbt-metabase models` with auto config loading:

image

An example config:

```yaml
config:
  dbt_database: reporting
  mb_user: alex@email
  ...
  mb_database: Reporting
```

## Type of change

- [x] New feature (non-breaking change which adds functionality)
- [x] This change requires a documentation update

# How Has This Been Tested?

Existing unit tests as well as production use.
In addition, this time around using `nox` I have tested the package on Python 3.6, 3.7 and 3.8

image

**Test Configuration**:

* Python version: 3.6
* Python version: 3.7
* Python version: 3.8

# Checklist:

- [x] My code follows the style guidelines of this project
- [x] I have performed a self-review of my own code
- [x] I have commented my code, particularly in hard-to-understand areas
- [x] I have made corresponding changes to the documentation
- [x] My changes generate no new warnings
- [x] I have added tests that prove my fix is effective or that my feature works
- [x] New and existing unit tests pass locally with my changes

### Closing Notes

Env Vars taking precedence over config.yml defaults (via injection)
CLI flags taking highest precedence as in a standard CLI tool (overrides)


z3z1ma:
@gouline 

Ready for review. This one is a good one.

README.rst
```diff
@@ -322,7 +322,7 @@ line. But if you prefer to call it from your code, here's how to do it:
     # Collect Args the Build Configs #
     ##################################
 
-    metabase_config = MetabaseConfig(
+    metabase_config = Metabase(
```
gouline, line 325: What's the thinking behind this? Config suffix described the intent of the object well.

dbtmetabase/logger/logging.py
```diff
@@ -0,0 +1,68 @@
...
+    level: str,
+) -> logging.StreamHandler:
+    """This handler routes logging output to the console but we are using rich instead"""
+    handler = logging.StreamHandler(sys.stdout)
+    handler.setFormatter(LOG_FILE_FORMAT)
+    handler.setLevel(getattr(logging, level.upper()))
+    return handler
+
+
+@lru_cache(maxsize=128)
```
gouline, line 43: Is the point of caching this, instead of using a static instance, just so the logger name can be applied? Level seems to be statically defined (intentionally or accidentally, considering that it's listed in arguments), so if it's just the name, the function with a cached return seems like an overkill.

You're also not setting any arguments in this project, so it's never going to change from default values.

dbtmetabase/models/interface.py
```diff
@@ -0,0 +1,165 @@
...
+        # Metabase additional connection opts
+        self.use_http = use_http
+        self.verify = verify
+        # Metabase Sync
+        self.sync = sync
+        self.sync_timeout = sync_timeout
+        self._client: Optional["MetabaseClient"] = None
+
+    @property
+    def client(self) -> "MetabaseClient":
```
gouline, line 44: Should the return type be in double quotes?

dbtmetabase/parsers/dbt_folder.py
```diff
@@ -1,12 +1,13 @@
-import logging
 import os
 import re
 import yaml
 from pathlib import Path
-from typing import List, Iterable, Mapping, MutableMapping, Optional
+from typing import List, Iterable, Mapping, MutableMapping, Optional, Tuple
 
-from ..models.metabase import METABASE_META_FIELDS, ModelKey
+from ..models import interface  # pylint: disable=unused-import
```
gouline, line 7: Why is this imported?

gouline:
Looks cool. Was this trying to solve any particular issue or just overall user friendliness?

We should really start stabilising the 0.8.0 interface and documentation, my concern is that we're up to RC4 and we're still making breaking changes...

z3z1ma:
Yes this one was fully focused on user friendliness. 

For the end user there is actually no change in this PR unless they are invoking the package through python in which case it definitely would be a breaking change. I was highly intentful on making the package work identically with the refactored CLI as it did before. 

The feature addition adds a new `config` command which lets users interactively generate a config which essentially caches arguments they would have to repeat on most invocations. Its totally optional for the end user though. In addition we support env vars but those are also completely optional. 

AFAIK we are only adding enhancements and no breaking changes.

README.rst
z3z1ma: Yeah I extended the class to be more generic and integrated the clients/parsers directly into the config. So because the classes were multipurpose, config and interface, I renamed to a more generic name. 
BUT what would be the best possible go at it is an interface class which inherits from the DbtConfig / MetabaseConfig
So I can definitely make that change. That way we keep the configs as base classes and extend them to the interface.

dbtmetabase/logger/logging.py
z3z1ma: Level is defined so we dont have to pass the value every time but the logger is able to be used with a passed level or different passed name in a specific call without creating multiple logger objects in your code.
Your right that we aren't logging with enough complexity to warrant, so we can definitely simplify this.

gouline:
That makes sense. For breaking changes, I meant if they're invoking it programmatically in Python.

Once this PR is merged, let's not have any more breaking changes for 0.8.x. I will release RC5, we can wait a few weeks to allow for finding/fixing any bugs and then I will release the final 0.8.0. Other features, such as metrics parsing, can go into 0.9.0 alpha/beta builds.

z3z1ma:
You got it. I appreciate your time and guidance.

dbtmetabase/logger/logging.py
z3z1ma: I parameterized it further and made the code in the logger more understandable. I went to remove the cache but that involved instantiating logger pointers in every module which started to deviate from the original intent; which was to unify and centralize logging to a single functional interface-- modifying logger module attributes anytime during execution for global overrides or providing args for locally scoped overrides on a per call basis. Neither of which can be done with a pointer without mutating the logger handlers locally. 

I know we don't use it for anything too complex today, but complexity isn't the issue -- using the cache here (I downsized it) and using the logger from the single interface `logger().info...` is technically _less_ code than it would be instantiating pointers in each module with logging.getLogger while being more powerful/flexible interface for the future of whoever contributes.

z3z1ma:
Alright.

I reinstated the config objects which can be used as they were and moved the interfaces to a separate module. The interfaces now inherit directly from the configs. This object inheritance is much more standard and clean practice. This makes it extremely clear that the interfaces extend configs with the addition of class methods and props that integrate their respective parsers/client without, as a user in Python, needing to build then shuttle the config between the other classes. The interface will give you access through props to a resolved parser, or client, perform validation, etc. The higher level things that weren't in scope of a config class alone.

The whole codebase is a treat to digest right now I think. But I defer to others to see if they share the sentiment 😄 

I think there is just documentation updates and anything else that can come from your review before its buttoned up and we leave it to stew.

README.rst
gouline: I think you forgot to reinstate this change in the README.

gouline:
Love your work. Looks good overall, I think it's ready for the documentation updates and then I'll do the final pass review before merging.

z3z1ma:
@gouline 

Documentation updates complete including the config section for end users too.

dbtmetabase/models/interface.py
```diff
@@ -0,0 +1,127 @@
...
+
+    @property
+    def parser(self) -> Union[DbtManifestReader, DbtFolderReader]:
+        if self._parser is None:
+            raise DbtParserNotInstantiated(
+                "dbt reader is not yet instantiated. Call `prepare_dbt_parser` method first"
+            )
+        return self._parser
+
+    def get_config(self) -> DbtConfig:
```
gouline, line 83: If `DbtInterface` class extends `DbtConfig`, you can just use the former as the latter. Or am I missing something?

z3z1ma: Yes but I though instantiating an interface to expose a parser and then passing the object (self) reference to that parser seemed a little circular. If I were going to do that, I'd just make a method that internally uses the parser and passes the self ref. 

This way is more verbose, but makes it clearer that the config (a DbtConfig) is all that is required to use the function --not the DbtInterface...
Does that make sense, what do you think?

gouline: I'd do the internal passing thing. Otherwise, that single use of it looks confusing - you're basically passing a representation of something to a function of a member of that something. That member (parser) should just be set up with a reference back to its owner (interface). 

For lifecycle purposes, interface still owns parser, but parser is just made aware of what owns it. Similar to how a child node in a graph implementation might have a reference back to its parent, in case it needs to propagate back up for any reason.

z3z1ma: Ok I added a method to the interface which passes its self ref to the function and users can use that instead of accessing the raw uninformed parser.
Definitely looks cleaner and makes more sense. Thank you

gouline (approved):
Love your work, thanks for your efforts, as always! 🚀

PR #70, opened 2021-10-13
Fix Docs and default value for sync timeout

### LINKED ISSUE #69: Doc update / default value needed: schema sync behavior
The sync timeout should have a default value otherwise the behavior is not as described
### Current behavior:
Sync is only attempted if metabase_sync_timeout is defined (default is None) 

### current docstring:
`metabase_sync_timeout (Optional[int], optional): Synchronization timeout (in secs). If set, we will fail hard on synchronization failure; if not set, we will proceed after attempting sync regardless of success. Only valid if sync is enabled. Defaults to None.`

https://github.com/gouline/dbt-metabase/blob/e7c7832c473b8d17ad2dd02f402c881e93897332/dbtmetabase/__init__.py#L546

mmmatthew:
Resolves #69

### DIFF
--- a/dbtmetabase/__init__.py
+++ b/dbtmetabase/__init__.py
@@ -520,7 +520,7 @@ def models(
     metabase_use_http: bool = False,
     metabase_verify: Optional[str] = None,
     metabase_sync: bool = True,
-    metabase_sync_timeout: Optional[int] = None,
+    metabase_sync_timeout: Optional[int] = 60,
     dbt_include_tags: bool = True,
     dbt_docs_url: Optional[str] = None,
     verbose: bool = False,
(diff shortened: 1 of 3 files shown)


z3z1ma:
It would look like this (not sure if we need an extra parens set around the first bool, don't think so)

```python

        # Sync and attempt schema alignment prior to execution; if timeout is not explicitly set, proceed regardless of success
        if self.sync:
            # If `sync_and_wait` returns True then regardless of timeout we proceed
            # If `sync_and_wait` returns false, the short circuit operator evals `self.sync_timeout is not None` before deciding to fail fast
            if not self._client.sync_and_wait(
                self.database,
                dbt_models,
                self.sync_timeout,
            ) and self.sync_timeout is not None:
                logging.critical("Sync timeout reached, models still not compatible")
                raise MetabaseUnableToSync(
                    "Unable to align models between dbt target models and Metabase"
                )

```

gouline:
Where is this at @z3z1ma @mmmatthew ? I'm cleaning up issues before a release.

z3z1ma:
@gouline 

I like what mmmathew did here. I think its a little cleaner/more explicit. Will wait to see if he has any more updates but looks good. Just need CI to pass.

mmmatthew:
Sorry, I forgot to answer here. I don't have any other updates but I can try to finish this next week

mmmatthew:
@gouline can you approve this?

gouline:
CI still failing, otherwise looks good.

gouline (approved):

PR #76, opened 2021-11-04
Feature/caveats and points of interest

patygold:
## Problem
There are plus two fields on Metabase table's description, which can also be added to this automation. On Metabase they are called “Why this table is interesting“ and “Things to be aware of about this table“.

## Approach
- Added parameters `points_of_interest` and `caveats` to the API call of table (https://www.metabase.com/docs/latest/api-documentation.html#put-apitableid)
- Adjusted README with those two new parameters

### DIFF
--- a/README.rst
+++ b/README.rst
@@ -67,6 +67,8 @@ Let's start by defining a short sample ``schema.yml`` as below.
     models:
       - name: stg_users
         description: User records.
+        points_of_interest: Relevant records.
+        caveats: Sensitive information about users.
         columns:
           - name: id
             description: Primary key.
--- a/README.rst
+++ b/README.rst
@@ -67,8 +67,6 @@ Let's start by defining a short sample ``schema.yml`` as below.
     models:
       - name: stg_users
         description: User records.
-        points_of_interest: Relevant records.
-        caveats: Sensitive information about users.
         columns:
           - name: id
             description: Primary key.
@@ -297,6 +295,24 @@ Here are the visibility types supported by Metabase:
 * ``retired`` (supported but not reflected in the UI)
 
 If you notice new ones, please submit a PR to update this readme.
+
+Model extra fields
+----------------
+
+In addition to the model description, metabase accepts two extra information fields. Those optional
+fields are called ``caveats`` and ``points_of_interest`` and can be defined under the ``meta`` tag
+of the model.
+
+This is how you can specify them in the ``stg_users`` example:
+
+.. code-block:: yaml
+
+  - name: stg_users
+    description: User records.
+    meta:
+      metabase.points_of_interest: Relevant records.
+      metabase.caveats: Sensitive information about users.
+
 
 Database Sync
 -------------
--- a/README.rst
+++ b/README.rst
@@ -299,7 +299,7 @@ If you notice new ones, please submit a PR to update this readme.
 Model extra fields
 ----------------
 
-In addition to the model description, metabase accepts two extra information fields. Those optional
+In addition to the model description, Metabase accepts two extra information fields. Those optional
 fields are called ``caveats`` and ``points_of_interest`` and can be defined under the ``meta`` tag
 of the model.
 
--- a/dbtmetabase/metabase.py
+++ b/dbtmetabase/metabase.py
@@ -220,14 +220,22 @@ class MetabaseClient:
             model_description = None
         else:
             model_description = model.description
+        if not model.points_of_interest:
+            model_points_of_interest = None
+        else:
+            model_points_of_interest = model.points_of_interest
+        if not model.caveats:
+            model_caveats = None
+        else:
+            model_caveats = model.caveats
 
         body_table = {}
         if api_table["description"] != model_description:
             body_table["description"] = model_description
-        if api_table["points_of_interest"] != model.points_of_interest:
-            body_table["points_of_interest"] = model.points_of_interest
-        if api_table["caveats"] != model.caveats:
-            body_table["caveats"] = model.caveats
+        if api_table.get("points_of_interest") != model_points_of_interest:
+            body_table["points_of_interest"] = model_points_of_interest
+        if api_table.get("caveats") != model_caveats:
+            body_table["caveats"] = model_caveats
 
         table_id = api_table["id"]
         if bool(body_table):
--- a/dbtmetabase/metabase.py
+++ b/dbtmetabase/metabase.py
@@ -222,12 +222,20 @@ class MetabaseClient:
             model_description = model.description
 
         table_id = api_table["id"]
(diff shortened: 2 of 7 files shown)


README.rst
```diff
@@ -67,6 +67,8 @@ Let's start by defining a short sample ``schema.yml`` as below.
     models:
       - name: stg_users
         description: User records.
+        points_of_interest: Relevant records.
```
gouline, line 70: These aren't in the dbt schema, you cannot add any custom fields in there. https://docs.getdbt.com/reference/configs-and-properties

gouline (changes requested):
These two new fields appear to only be supported in Metabase? They are not in the dbt schema, you have to provide them in the `meta` tags, similar to what we do with `special_types`.

README.rst
patygold: @gouline Thanks! I am adjusting this point. One follow-up question: Is this a best practice of this repository to start these meta definitions with `metabase.`, as the `special_type` is currently defined?

gouline: That's correct. Anything can be added to `meta`, so to differentiate fields relating to Metabase, we prefix everything with `metabase.`.

dbtmetabase/parsers/dbt_folder.py
```diff
@@ -144,6 +144,9 @@ def _read_model(
             metabase_columns.append(self._read_column(column, schema))
 
         description = model.get("description", "")
+        meta = model.get("meta")
```
gouline, line 147: If you retrieve it as `model.get("meta", {})`, you don't need to do ternaries for each field below.

dbtmetabase/metabase.py
```diff
@@ -222,12 +222,20 @@ def export_model(
...
+                api_table["points_of_interest"] != model.points_of_interest or
+                api_table["caveats"] != model.caveats
+        ):
             # Update with new values
             self.api(
                 "put",
                 f"/api/table/{table_id}",
-                json={"description": model_description},
+                json={
+                    "description": model_description,
```
gouline, line 235: Only fields that are different should be persisted, not all 3 fields if at least one of them is different.

gouline (changes requested):
See inline comments. You only implemented the dbt_folder parser (soon to be deprecated), you need to implement dbt_manifest as well. 

Finally, CI validation is failing, please make sure `make check-fmt check-lint check-type` is passing locally before committing. Here's what the GitHub Action is running https://github.com/gouline/dbt-metabase/blob/master/.github/workflows/pull_request.yml#L25-L41.

dbtmetabase/metabase.py
patygold: Adjusted 👍

patygold:
@gouline BTW, I've run the `check-type` locally, and the error seems not to be related to my changes specifically.
Maybe you have more information on that? Specifically, the error I get is:
````
(agr-meltano) ➜  dbt-metabase git:(feature/caveats_and_points_of_interest) ✗ make check-fmt check-lint check-type
black --check .
All done! ✨ 🍰 ✨
18 files would be left unchanged.
pylint dbtmetabase

--------------------------------------------------------------------
Your code has been rated at 10.00/10 (previous run: 10.00/10, +0.00)

mypy dbtmetabase
dbtmetabase/__init__.py:7: error: Skipping analyzing "click": found module but no type hints or library stubs
dbtmetabase/__init__.py:7: note: See https://mypy.readthedocs.io/en/stable/running_mypy.html#missing-imports
Found 1 error in 1 file (checked 14 source files)
make: *** [check-type] Error 1

**Comment by gouline** [comment]
> @gouline BTW, I've run the `check-type` locally, and the error seems not to be related to my changes specifically. Maybe you have more information on that? Specifically, the error I get is:

GitHub Action ran successfully, so it must be your local setup. I'll try to review your changes this week (or @z3z1ma tag in, if you have time).

**Comment by z3z1ma** [comment]
:+1:   I should have some time tomorrow to jump in and review.

**Comment by patygold** [comment]
@z3z1ma Any news with this review?

File: dbtmetabase/metabase.py [hunk 4]
```diff
@@ -221,13 +221,21 @@ def export_model(
         else:
             model_description = model.description
 
+        body_table = {}
+        if api_table["description"] != model_description:
+            body_table["description"] = model_description
+        if api_table["points_of_interest"] != model.points_of_interest:
```
gouline, line 227: Do these get returned if they haven't been set for all versions of Metabase? Just making sure it doesn't blow up for anyone using older Metabase versions. If not, then checking for `"points_of_interest" in api_table` first might be useful.

README.rst
```diff
@@ -296,6 +296,24 @@ Here are the visibility types supported by Metabase:
 
 If you notice new ones, please submit a PR to update this readme.
 
+Model extra fields
+----------------
+
+In addition to the model description, metabase accepts two extra information fields. Those optional
```
gouline, line 302: Capitalise "Metabase" please.

gouline (commented):
Not sure about @z3z1ma, but I'm pretty happy with it other than few minor things I noted.

The only remaining item is adding unit tests.

dbtmetabase/metabase.py
z3z1ma: Agree, better to lean safe here with a `get` accessor.

z3z1ma:
Sorry @patygold @gouline 
Work has been nuts lately. 

It all looks good to me, just one comment I on the possible None type string being passed into API which I believe is rejected.

dbtmetabase/metabase.py
```diff
@@ -221,13 +221,21 @@ def export_model(
...
+            body_table["points_of_interest"] = model.points_of_interest
+        if api_table["caveats"] != model.caveats:
```
z3z1ma, line 229: We have a scenario here where the user has existing `caveats` defined, then removes it before the next run through. In this case api_table["caveats"] is not equal to the now Nonetype parsed caveats which will trigger an update. This update will pass the key caveats with None as the value. Metabase would definitely reject that if it were the description field so I would wager it might for the caveats / points of interest. We may need to set the dataclass up so that it doesn't allow None and default to empty str instead.

Its something we should test (verify it on a real metabase instance or in their documentation) and account for this possibility.

dbtmetabase/metabase.py
patygold: Nice! I've adjusted to use `get()` 👍

dbtmetabase/metabase.py
patygold: Ah! True. I've basically adjusted this part, based on the `model_description` logic.

gouline (approved):
Looks good, thanks for making these changes!

Leaving this for a day or so for @z3z1ma to comment with any concerns, otherwise merging.

PR #83, opened 2022-01-12
feat: Support --metabase_session_id

lawrencejones:
Metabase lacks proper support for authenticating from automation, which
means we have to exchange a username and password for a session token
before each invocation of this command.

These count as login attempts, which are rate limited. Depending on your
needs, you may hit the rate limit and find your schemas can no longer be
synced.

Providing a --metabase_session_id allows the user to reuse and existing
Metabase session, should they have it, allowing them to avoid the login
rate limits (and in some cases, an annoying email!).


lawrencejones:
Hey team! I haven't written any Python in years so this is unlikely to work, but I'd love early feedback on the proposed change.

I'll make sure this is actually functional once it gets the green light.

gouline:
Approach looks fine overall, you have some syntax errors as the CI shows. Please run these checks locally to debug and let me know when you're ready for a code review.

lawrencejones:
Yep, I definitely eye balled this! Now I know you're ok with the approach I'll fix it up on Monday :)

gouline:
@lawrencejones Still planning to get this merged?

Issue #84, opened 2022-01-13
Better Documentation


soltanianalytics:
I'd wager that the usability and prevalence of this marvelous packages would be greatly advanced if there was a basic documentation to go along with the two major models (`MetabaseInterface` and `DbtInterface`). For instance, it took me a few hours to realize that my implementation wasn't working because `schema` is, in fact, not a mandatory field but rather a field limiting the amount of models being parsed and synced.

I think this is a great tool that is easily usable _if you know how to_, where a bit of docs could go a long way :)

gouline:
Feel free to propose specific improvements. We went as far as we could with the documentation, beyond that you're welcome to PR improvements/corrections.

soltanianalytics:
Solved

PR #85, opened 2022-01-13
Add interface kwarg docs to README.rst

soltanianalytics:
PR relating to https://github.com/gouline/dbt-metabase/issues/84

### DIFF
--- a/README.rst
+++ b/README.rst
@@ -135,7 +135,7 @@ how our dbt models are exposed in BI which closes the loop between ELT, modellin
         --output_path ./models/ \
         --output_name metabase_exposures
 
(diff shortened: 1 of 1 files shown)


gouline:
This will become outdated almost immediately. I would prefer if you created [PEP 257](https://www.python.org/dev/peps/pep-0257/)-compliant docstrings in `DbtConfig` and `MetabaseConfig` constructors explaining what each field means (you can find meanings in help messages in the root [__init__.py](https://github.com/gouline/dbt-metabase/blob/master/dbtmetabase/__init__.py)).

gouline:
Superseded by #87

PR #87, opened 2022-01-14
Merge config into interface to avoid confusion

gouline:
Seeing confusion in #84 and #85, this simplifies things.


gouline:
@z3z1ma What do you think? This simplifies programmatic usage as much as possible without breaking changes in any documented APIs (anyone using undocumented stuff should be capable of updating).

@soltanianalytics Would this make your life easier? Everything now has a docstring visible in your IDE.

soltanianalytics:
Yes! Thanks a lot 👍

dbtmetabase/models/interface.py
```diff
@@ -37,6 +67,11 @@ def prepare_metabase_client(self, dbt_models: Optional[List[MetabaseModel]] = No
         Raises:
             MetabaseUnableToSync: This error is raised if sync is enabled and a timeout is explicitly set in the `Metabase` object config
         """
+
+        if self._client is not None:
```
z3z1ma, line 71: Good use of a singleton pattern, should `prepare_metabase_client` just be private at this point? ie: `_prepare_metabase_client`, I don't think further encapsulation would affect usability honestly, its pretty straightforward.

dbtmetabase/parsers/dbt.py
```diff
@@ -0,0 +1,46 @@
...
+    def __init__(
+        self,
+        path: str,
+        database: str,
+        schema: Optional[str],
+        schema_excludes: Optional[Iterable],
+        includes: Optional[Iterable],
+        excludes: Optional[Iterable],
+    ):
+        """[summary]
```
z3z1ma, line 20: should this still say [summary] ?

z3z1ma (dismissed):
This look great. I like the abc class in the parsers and generally everything feels more intuitive, both from an extensibility and usability perspective.

dbtmetabase/parsers/dbt.py
gouline: Oops, good catch!

dbtmetabase/models/interface.py
gouline: Agreed, I kept it for backwards compatibility because it's called out in the current readme. We can make it private in 0.9.x in the next round of changes.

PR #95, opened 2022-04-12
feat: Add flag to skip syncing sources, and allow meta-defined display names

atvaccaro:
This tool is amazing!

I have a couple features that may be helpful.

First, adding a flag to skip all dbt sources. In [our project](https://github.com/cal-itp/data-infra/tree/main/warehouse), we ultimately use Metabase on top of BigQuery, but we don't import source datasets into Metabase. Without the flag, `dbt-metabase` throws errors as it tries to sync dbt source models.

Second, adding `metabase.display_name` as an optional field. We deal with a lot of [GTFS](https://gtfs.org/) data, so our model names naturally have GTFS as a common acronym. Currently it gets title-cased as `Gtfs` in many situations, which is annoying. So I wanted to have the ability to override the Metabase `display_name` property.

### DIFF
--- a/dbtmetabase/__init__.py
+++ b/dbtmetabase/__init__.py
@@ -500,6 +500,11 @@ def config(ctx, inspect: bool = False, resolve: bool = False, env: bool = False)
     help="Flag to append tags to table descriptions in Metabase (default False)",
 )
 @click.option(
+    "--metabase_sync_sources",
+    is_flag=True,
+    help="If True, try to sync dbt sources to Metabase. Defaults to False.",
+)
+@click.option(
     "-v",
     "--verbose",
     is_flag=True,
@@ -520,6 +525,7 @@ def models(
     metabase_use_http: bool = False,
     metabase_verify: Optional[str] = None,
     metabase_sync: bool = True,
+    metabase_sync_sources: bool = False,
     metabase_sync_timeout: Optional[int] = None,
     dbt_include_tags: bool = True,
     dbt_docs_url: Optional[str] = None,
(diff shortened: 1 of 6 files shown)


dbtmetabase/logger/logging.py
```diff
@@ -1,4 +1,5 @@
 import logging
+import os
```
gouline, line 2: Redundant.

dbtmetabase/__init__.py
```diff
@@ -520,6 +525,7 @@ def models(
     metabase_use_http: bool = False,
     metabase_verify: Optional[str] = None,
     metabase_sync: bool = True,
+    metabase_sync_sources: bool = False,
```
gouline, line 528: "Sync" only relates to polling for changes, not for the action of _exporting_ the models. Your change also changes the default behaviour, which will break the build for anyone expecting the sources to be exported.

The flag should be something along the lines of `exclude_sources` default to false and only NOT export sources when it's set to true.

gouline (changes requested):
Run CI checks locally and fix any failures _before_ pushing, GitHub waits for me to run them in Actions for first-time contributors.

@z3z1ma Do you remember what source models are used for after reading, apart from being exported to Metabase? If they are not exported, is there any point in reading them in the first place? Trying to think whether they should be excluded from the Metabase client (how this PR implements it) or from being read in the parsers.

atvaccaro:
> Run CI checks locally and fix any failures _before_ pushing, GitHub waits for me to run them in Actions for first-time contributors.

Ah I'm sorry! To be honest I didn't notice them until I got the failure email, but will do that.

Re: formatting, I don't see a `black` version pinned. Would you be open to pinning 22.3.0?

z3z1ma:
@atvaccaro 

black is in `requirements-test.txt` but it isn't pinned. No deps are pinned currently actually, though the latest compatible versions would be chosen on fresh install.

@gouline 
`dbt-metabase models`  only uses source for exporting definitions.
`dbt-metabase exposures` uses sources as possible matches for BI exposure extraction

I think it might be cleaner/more logical to exclude in the parser with a flag and only pass that flag from the `models` command when the CLI flag `exclude_sources` is passed

gouline:
> Re: formatting, I don't see a `black` version pinned. Would you be open to pinning 22.3.0?

No regular work is done on this repository so nobody will be updating those pinned versions, use latest. Create a virtual env if it conflicts with other projects.

gouline:
@atvaccaro Still planning to get this merged?

atvaccaro:
> @atvaccaro Still planning to get this merged?

Hey I'm so sorry and thank you for getting this through. I've been heads-down on a major project and refactor and never got back to this.

Issue #88, opened 2022-02-02
Deactiveted user in metabase causes an error during exposures sync


arathunku:
Hello,

when a user is deactivated, metabase returns 404 for any API requests regarding this user, and this causes an exception during exposures sync:

```
Traceback (most recent call last):
  File "/usr/local/lib/python3.8/site-packages/dbtmetabase/metabase.py", line 847, in api
    response.raise_for_status()
  File "/usr/local/lib/python3.8/site-packages/requests/models.py", line 940, in raise_for_status
    raise HTTPError(http_error_msg, response=self)
requests.exceptions.HTTPError: 404 Client Error: Not Found for url: https://metabase-internal.gatserver.com/api/user/15
During handling of the above exception, another exception occurred:
Traceback (most recent call last):
  File "/usr/local/bin/dbt-metabase", line 5, in <module>
    dbtmetabase.main()
  File "/usr/local/lib/python3.8/site-packages/dbtmetabase/__init__.py", line 720, in main
    cli(max_content_width=600)  # pylint: disable=unexpected-keyword-arg
  File "/usr/local/lib/python3.8/site-packages/click/core.py", line 1128, in __call__
    return self.main(*args, **kwargs)
  File "/usr/local/lib/python3.8/site-packages/click/core.py", line 1053, in main
    rv = self.invoke(ctx)
  File "/usr/local/lib/python3.8/site-packages/click/core.py", line 1659, in invoke
    return _process_result(sub_ctx.command.invoke(sub_ctx))
  File "/usr/local/lib/python3.8/site-packages/dbtmetabase/__init__.py", line 113, in invoke
    return super().invoke(ctx)
  File "/usr/local/lib/python3.8/site-packages/click/core.py", line 1395, in invoke
    return ctx.invoke(self.callback, **ctx.params)
  File "/usr/local/lib/python3.8/site-packages/click/core.py", line 754, in invoke
    return __callback(*args, **kwargs)
  File "/usr/local/lib/python3.8/site-packages/dbtmetabase/__init__.py", line 262, in wrapper
    return func(*args, **kwargs)
  File "/usr/local/lib/python3.8/site-packages/dbtmetabase/__init__.py", line 709, in exposures
    metabase.client.extract_exposures(
  File "/usr/local/lib/python3.8/site-packages/dbtmetabase/metabase.py", line 583, in extract_exposures
    creator = self.api("get", f"/api/user/{exposure['creator_id']}")
[...]

PR #96, opened 2022-04-19
fix: Handle 404s for deactivated users when exporting exposures

kochalex:
This fixes an issue where exporting exposures crashes `dbt-metabase` when the owner of a card is a deactivated user. The metabase API has a check if a user is deactivated when running a GET to `/api/user/:id` and returns a 404.

Fixes: https://github.com/gouline/dbt-metabase/issues/88


dbtmetabase/metabase.py
```diff
@@ -819,6 +823,10 @@ def _build_exposure(
                 if exposure.upper() in refable_models
             ],
         }
+        if creator_email is None:
+            del exposure["owner"]
```
kochalex, line 827: Deleting the key if the variable is not defined works around a type error that came up when doing the more obvious approach of setting the key if the variable is defined. There could be a more elegant solution here...

tests/fixtures/mock_api/api/card/3.json
```diff
@@ -1 +1 @@
...
...global": {"distinct-count": 1, "nil%": 0.0}, "type": {"type/Number": {"min": 100.0, "q1": 100.0, "q3": 100.0, "max": 100.0, "sd": null, "avg": 100.0}}}}], "can_write": true, "database_id": 2, "enable_embedding": false, "collection_id": 3, "query_type": "query", "name": "Total customers", "dashboard_count": 1, "creator_id": 2, "updated_at": "2021-07-21T08:01:41.996Z", "made_public_by_id": null, "embedding_params": null, "cache_ttl": null, "dataset_query": {"type": "query", "database": 2, "query": {"source-table": 7, "aggregation": [["count"]]}}, "id": 3, "display": "scalar", "last-edit-info": {"id": 1, "email": "user@example.com", "first_name": "Alexander", "last_name": "Butler", "timestamp": "2021-07-21T08:01:37.449936Z"}, "visualization_settings": {"graph.series_labels": ["number"], "graph.metrics": ["count"], "graph.dimensions": []}, "collection": {"authority_level": null, "description": "Automatically generated cards.", "archived": false, "slug": "a_look_at_your_customers_table", "color": "#f9d45c", "name": "A look at your customers table", "personal_owner_id": null, "id": 3, "location": "/2/", "namespace": null}, "created_at": "2021-07-21T08:01:37.434243Z", "public_uuid": null}
```
kochalex, line 1: This was changed to exercise the modified code by removing the "creator" key, causing it to look up the creator via id, which does not exist.

dbtmetabase/metabase.py
```diff
@@ -593,9 +593,12 @@ def increase_indent(self, flow=False, indentless=False):
                     creator_email = exposure["creator"]["email"]
                     creator_name = exposure["creator"]["common_name"]
                 elif "creator_id" in exposure:
-                    creator = self.api("get", f"/api/user/{exposure['creator_id']}")
-                    creator_email = creator["email"]
-                    creator_name = creator["common_name"]
+                    # If a metabase user is deactivated, the API returns a 404
+                    creator = self.api(
```
gouline, line 597: When critical is set to false, it allows any response (not just 404). Ideally, it should check specifically for 404 that's allowed.

tests/fixtures/mock_api/api/card/3.json
gouline: @z3z1ma You wrote these tests, can you have a look to make sure you're cool with this?

dbtmetabase/metabase.py
```diff
@@ -859,12 +868,12 @@ def api(
             try:
                 response.raise_for_status()
             except requests.exceptions.HTTPError:
-                if "password" in kwargs["json"]:
+                if "json" in kwargs and "password" in kwargs["json"]:
```
kochalex, line 871: At least for the error I was getting with exposures, kwargs does not have "json", it is just the auth token for metabase.

kochalex:
Updated the error handling - I'm also happy to add a new card to the mocked responses to test the 404 per your other comment if you would like.

dbtmetabase/metabase.py
z3z1ma: @kochalex  Is this related to the same issue as the ticket?

tests/fixtures/mock_api/api/card/3.json
z3z1ma: @kochalex I created the test mocks using a real scraped metabase instance so I cant tell at a glance but can you just verify the json is modified in such a way that it reflects what metabase returns in the test case we are recreating?

To just delete the owner key but that not be a reflection of reality might impact future development in an unexpected way causing tests that fail when they are valid against metabase API. 

Just checking 🪂

kochalex: This does cause the same error that occurs against a real metabase instance. I'll add an additional card mock API reponse though and try to export it directly from a test metabase deployment.

Here's what was happening in the error cases when running dbt-metabase against on our metabase deployment:
* Metabase's API returns cards that look like this - with only a creator_id but not the creator details.
  *  This is the case for at least some cards where the owner has been deleted (it may be for others as well)
* When dbt-metabase get a card response where there is a creator_id but not the creator details in the response, it makes a call to metabase's user API
  * Metabase's user API filters out deleted users and you are not able to retrieve information on them.
  * dbt-metabase crashes as this point

dbtmetabase/metabase.py
kochalex: It is - the error handling here errors out when dbt-metabase makes a GET for a deleted user when exporting exposures.

That said, I did not dig to see if this was specific to GET "/api/user/deleted_user_id" or a more general case for other API endpoints - I intended for this change to "do no harm"

z3z1ma:
I'm good with this pending the additional test mock plus making CI pass.

gouline (approved):

PR #100, opened 2022-05-01
Ensures that boolean option flags are overridden by config.yml when no CLI flags are provided

### LINKED ISSUE #99: metabase_use_http argument not being used when specified in config.yml
When using a config.yml file and specifying `metabase_use_http: True` (tried upper and lower case), I end up with the following SSL error (**note: running Metabase via `http://localhost:3000`**):

```
requests.exceptions.SSLError: HTTPSConnectionPool(host='localhost', port=3000): Max retries exceeded with url: /api/session (Caused by SSLError(SSLError(1, '[SSL: WRONG_VERSION_NUMBER] wrong version number (_ssl.c:1122)')))
```

After digging into it, the issue was that it was still sending the request via `https`. I found that [here](https://github.com/gouline/dbt-metabase/blob/master/dbtmetabase/metabase.py#L854-L856) the `self.protocol` attribute was still being passed as `https` resulting in the error.

```
response = requests.request(
            method, f"{self.protocol}://{self.host}{path}", verify=self.verify, **kwargs
        )
```

Using --metabase_http as a command line argument fixed the issue for me, I _believe_ it just overrides what's [here](https://github.com/gouline/dbt-metabase/blob/master/dbtmetabase/__init__.py#L236-L241)
```
@click.option(
    "--metabase_http/--metabase_https",
    "metabase_use_http",
    default=False,
    help="use HTTP or HTTPS to connect to Metabase. Default HTTPS",
)
```

Currently using dbt-metabase==0.8.5

jack-cook-repo:
Previously, any values passed in a config.yml file for boolean flags (metabase_use_http, metabase_sync) were ignored.

This change will override any flags that are of type click.types.BoolParamType with values provided in a config.yml file, even if specified in a command line argument.

For example, when `metabase_use_http` is set to `True` in config.yml, running `dbt-metabase models --metabase_https` will still default to using HTTP.

Resolves #99

### DIFF
--- a/dbtmetabase/__init__.py
+++ b/dbtmetabase/__init__.py
@@ -82,11 +82,17 @@ class ListParam(click.Tuple):
 
 class OptionAcceptableFromConfig(click.Option):
     """This class override should be used on arguments that are marked `required=True` in order to give them
-    more resilence to raising an error when the option exists in the users config"""
+    more resilence to raising an error when the option exists in the users config.
+
+    This also overrides boolean flags (e.g. --use_metabase_http/--use_metabase_https) in options
+    when a value is provided in the config.yml file. (e.g. metabase_use_http: True)"""
 
     def process_value(self, ctx: click.Context, value: Any) -> Any:
         if value is not None:
             value = self.type_cast_value(ctx, value)
+
+        if self.name in CONFIG and isinstance(self.type, click.types.BoolParamType):
+            value = CONFIG[self.name]
 
         if self.required and self.value_is_missing(value):
             if self.name not in CONFIG:
@@ -237,6 +243,7 @@ def shared_opts(func: Callable) -> Callable:
         "--metabase_http/--metabase_https",
         "metabase_use_http",
         default=False,
+        cls=OptionAcceptableFromConfig,
         help="use HTTP or HTTPS to connect to Metabase. Default HTTPS",
     )
     @click.option(
@@ -248,6 +255,7 @@ def shared_opts(func: Callable) -> Callable:
     @click.option(
         "--metabase_sync/--metabase_sync_skip",
         "metabase_sync",
+        cls=OptionAcceptableFromConfig,
         default=True,
         help="Attempt to synchronize Metabase schema with local models. Default sync",
     )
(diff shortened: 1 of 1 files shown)


gouline:
That's not an acceptable fix to the logic. Command-line flags should always override the config file, what you want is to use the config value when no command-line flag is passed.

So in your example, if `metabase_use_http` is set to `True` in config.yml, running `dbt-metabase models --metabase_https` should use HTTPS, but running `dbt-metabase models` should use HTTP.

z3z1ma:
Hey @jack-cook-repo  👋 

The precedence is:
1. CLI Flag
2. Env Var
3. config.yaml (consider these baseline defaults which can be overridden via either of the above methods)

Configuration is resolved in that order. 👆 

image

We are injecting config _ONLY_ if it 🗒️ :
  A) matches an existing CLI flag name in click context `ctx`
  B) no value was provided by the CLI flag  or via env var

So we should really dump that context var `ctx.params` seen in the screenshot above and verify what the CLI flag is called in the context.
Given those two steps, the solution should be (hopefully) straightforward

I am wondering if `metabase_https` ends up in the context instead of the switch flags target var name of `metabase_use_http`

jack-cook-repo:
Thanks both - I've pushed a change that checks, for boolean flags, whether `ctx._parameter_source[f'{self.name}']._name_ == 'DEFAULT'` (i.e., no CLI flag has been passed), **then** it checks to see whether a value was passed in the config file and overrides the default boolean value if so.

 **Testing**
No CLI flag, `metabase_use_http: True` specified in config.yml.
image

CLI flag that matches setting in config.yml
image

CLI flag that **doesn't** match setting in config.yml, and CLI flag takes precedence
image

This variable doesn't have an `envvar` set up so it just has preference of CLI flag > config.yml

z3z1ma (approved):
LGTM

z3z1ma:
Looks like it needs to be formatted with `black` though. 
Make sure to run CI stuff locally @jack-cook-repo to verify everything. The makefile can make it easier. Otherwise good

jack-cook-repo:
Thanks @z3z1ma - this is my first package contribution so I should have probably asked about the process! 

I ran `make` which appears to have run the Makefile without any issue - should I expected to have seen changes once `black` had run?

gouline:
CI seems to be stalled but I'll merge it in. Sorry for the delay, haven't been looking at this repo.

PR #111, opened 2022-07-06
Add `display_name` option for field names

jack-cook-repo:
Piggybacking off https://github.com/gouline/dbt-metabase/pull/109/files - add the ability to use display_name for fields

https://www.metabase.com/docs/latest/api/field.html#params-12


jack-cook-repo:
This is failing a test but I'm not sure which side (test file not set up correctly, or code to pull in the display_name isn't working).

Will give this another look at some point

PR #112, opened 2022-07-07
feat: Add `display_name` option for field names

gouline:
Fix and refactor for #111


jack-cook-repo:
Thank you for picking this up @gouline

PR #121, opened 2022-07-07
feat: allow self reference

### LINKED ISSUE #74: Add support for self-referencing tables
This logic does not account for self-referencing tables: 

https://github.com/gouline/dbt-metabase/blob/e7c7832c473b8d17ad2dd02f402c881e93897332/dbtmetabase/parsers/dbt_manifest.py#L209-L217

joelluijmes:
New approach of #110. This logic now allows the self reference / circular reference. 

I made this logic based on my initial work (i.e., before refactor of @gouline). By removing the set, it's now possible to have a self reference. In addition, as some assumptions are made I added some warning logger if in future dbt manifest changes.

Resolves #74

### DIFF
--- a/dbtmetabase/parsers/dbt_manifest.py
+++ b/dbtmetabase/parsers/dbt_manifest.py
@@ -201,12 +201,26 @@ class DbtManifestReader(DbtReader):
                 # would return the ref() written in the test, but if the model has an alias, that's not enough.
                 # It is better to use child['depends_on']['nodes'] and exclude the current model
 
-                depends_on_ids = set(child["depends_on"][model_type])
-                depends_on_ids.discard(model["unique_id"])
-                if not depends_on_ids:
+                # From experience, nodes contains at most two tables: the referenced model and the current model.
+                # Note, sometimes only the referenced model is returned.
+                depends_on_nodes = list(child["depends_on"][model_type])
+                if len(depends_on_nodes) > 2:
+                    logger().warning(f'Expected at most two nodes, got {len(depends_on_nodes)} nodes, skipping {model["unique_id"]}')
                     continue
 
-                depends_on_id = depends_on_ids.pop()
+                # Remove the current model from the list. Note, remove() only removes the first occurence. This ensures
+                # the logic also works for self referencing models.
+                try:
+                    depends_on_nodes.remove(model["unique_id"])
+                except KeyError:
+                    logger().warning(f'Expected nodes to contain current model, skipping {model["unique_id"]}')
+                    continue
+
+                if not any(depends_on_nodes):
+                    logger().warning(f'Expected single node after filtering, got {len(depends_on_nodes)} nodes, skipping {model["unique_id"]}')
+                    continue
+
+                depends_on_id = depends_on_nodes[0]
 
                 foreign_key_model = manifest[model_type].get(depends_on_id, {})
                 fk_target_table_alias = foreign_key_model.get(
--- a/dbtmetabase/parsers/dbt_manifest.py
+++ b/dbtmetabase/parsers/dbt_manifest.py
@@ -185,7 +185,9 @@ class DbtManifestReader(DbtReader):
 
         metabase_column: List[MetabaseColumn] = []
 
-        children = manifest["child_map"][model["unique_id"]]
+        unique_id = model["unique_id"]
+
+        children = manifest["child_map"][unique_id]
         relationship_tests = {}
 
         for child_id in children:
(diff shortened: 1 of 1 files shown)


dbtmetabase/parsers/dbt_manifest.py
```diff
@@ -201,12 +201,26 @@ def _read_model(
                 # would return the ref() written in the test, but if the model has an alias, that's not enough.
                 # It is better to use child['depends_on']['nodes'] and exclude the current model
 
-                depends_on_ids = set(child["depends_on"][model_type])
-                depends_on_ids.discard(model["unique_id"])
-                if not depends_on_ids:
+                # From experience, nodes contains at most two tables: the referenced model and the current model.
+                # Note, sometimes only the referenced model is returned.
+                depends_on_nodes = list(child["depends_on"][model_type])
```
gouline, line 206: Don't convert everything to list, use the API of whatever data structure (e.g. set) it's currently in.

dbtmetabase/parsers/dbt_manifest.py
```diff
@@ -201,12 +201,26 @@ def _read_model(
...
+                if len(depends_on_nodes) > 2:
+                    logger().warning(f'Expected at most two nodes, got {len(depends_on_nodes)} nodes, skipping {model["unique_id"]}')
                     continue
 
-                depends_on_id = depends_on_ids.pop()
+                # Remove the current model from the list. Note, remove() only removes the first occurence. This ensures
+                # the logic also works for self referencing models.
+                try:
+                    depends_on_nodes.remove(model["unique_id"])
+                except KeyError:
```
gouline, line 215: Check it's in there instead of catching exceptions.

dbtmetabase/parsers/dbt_manifest.py
```diff
@@ -201,12 +201,26 @@ def _read_model(
...
+                    logger().warning(f'Expected nodes to contain current model, skipping {model["unique_id"]}')
+                    continue
+
+                if not any(depends_on_nodes):
```
gouline, line 219: Redundant `any`.

dbtmetabase/parsers/dbt_manifest.py
```diff
@@ -201,12 +201,26 @@ def _read_model(
...
+                    logger().warning(f'Expected single node after filtering, got {len(depends_on_nodes)} nodes, skipping {model["unique_id"]}')
+                    continue
+
+                depends_on_id = depends_on_nodes[0]
```
gouline, line 223: Again, no need to convert it to a list to get the only item - that's exactly what `.pop()` on sets is for.

gouline (changes requested):
Your PRs are still failing formatter check, run `make check` and `make test` before pushing.

dbtmetabase/parsers/dbt_manifest.py
```diff
@@ -201,12 +201,26 @@ def _read_model(
...
                 # It is better to use child['depends_on']['nodes'] and exclude the current model
 
-                depends_on_ids = set(child["depends_on"][model_type])
-                depends_on_ids.discard(model["unique_id"])
-                if not depends_on_ids:
+                # From experience, nodes contains at most two tables: the referenced model and the current model.
+                # Note, sometimes only the referenced model is returned.
+                depends_on_nodes = list(child["depends_on"][model_type])
+                if len(depends_on_nodes) > 2:
+                    logger().warning(f'Expected at most two nodes, got {len(depends_on_nodes)} nodes, skipping {model["unique_id"]}')
```
gouline, line 208: Use lazy formatting (e.g. `%s`) instead of f-strings in logger statements.

dbtmetabase/parsers/dbt_manifest.py
```diff
@@ -201,12 +201,26 @@ def _read_model(
...
                     continue
 
-                depends_on_id = depends_on_ids.pop()
+                # Remove the current model from the list. Note, remove() only removes the first occurence. This ensures
+                # the logic also works for self referencing models.
+                try:
+                    depends_on_nodes.remove(model["unique_id"])
```
gouline, line 214: There's a line further down

```py
unique_id = model["unique_id"]
```

Move it to the top of the function so you don't have to extract it from the map every time.

dbtmetabase/parsers/dbt_manifest.py
joelluijmes: The current model is not always within the referenced nodes. Sometimes the list only contains the referenced node. I couldn't find any documentation how the nodes are structured, nor have the time to dig into the dbt source code to find out. 

Regardless, just checking if it is in it is thus invalid.

dbtmetabase/parsers/dbt_manifest.py
joelluijmes: Sure will adjust. Can I ask why? Or is this just your convention?

dbtmetabase/parsers/dbt_manifest.py
joelluijmes: I refactored it a bit. It now checks for length two, and then it should remove the current model.

dbtmetabase/parsers/dbt_manifest.py
gouline: It's a common convention (not just mine) in many languages, my IDE even highlights it as a warning. 

As I understand it, string formatting is relatively expensive, so you want to make it as lazy as possible by only processing it whenever that statement is above the log level (e.g. if your log level is set to warning, you don't want to waste time formatting all the info statements). We don't have a high-performance code here by any means but just a convention.

gouline:
> Yes this PR properly fixes #74. That is why I had to remove the set logic 🙂

Excellent, adding a resolve hook to the description 🎉

dbtmetabase/parsers/dbt_manifest.py
```diff
@@ -201,12 +203,31 @@ def _read_model(
...
+                    logger().warning(
+                        "Expected at most two nodes, got %d {} nodes, skipping %s {}",
+                        len(depends_on_nodes),
+                        unique_id,
+                    )
+                    continue
+
+                # Remove the current model from the list. Note, remove() only removes the first occurrence. This ensures
+                # the logic also works for self referencing models.
+                if len(depends_on_nodes) == 2 and unique_id in depends_on_nodes:
```
gouline, line 219: Do you need the `len == 2` condition here? You only care whether it's in there (and hence removing won't cause a KeyError), you're checking the length after that removal below for the ultimate return condition anyway.

joelluijmes: Yes I would say so. What if the nodes only contains the self referenced model, i.e., ‘len == 1’? In that case, we shouldn't remove it from the list.

gouline (approved):

PR #123, opened 2022-07-07
make dbt_includes and dbt_excludes case insensitive

### LINKED ISSUE #122: dbt_includes flag should be case insensitive
dbt model names are case insensitive, and so the user should be able to enter the names of included models in lowercase, which is often how models are typed.

Currently, because `model_name` in `dbt_manifest.py` is uppercased, model names that are sent to the `includes` variable as lowercase strings are ignored.

A simple uppercasing of the `includes` list solves this.

I'll open a PR.

_Note: I think a similar issue may exist with `dbt_excludes`, but I don't need to use that at the moment, and when I tried to include it as a quick fix, it didn't work expected and I don't have time to dig into it, so I'm leaving it for now._

leoebfolsom:
This PR addresses the issue I describe in #122 so that the user can enter a `--dbt_includes` argument with any casing (presumably lowercase will be most common) without conflicting with `model_name`, which is uppercase, in `parsers/dbt_manifest.py`.

Resolves #122

### DIFF
--- a/dbtmetabase/parsers/dbt_manifest.py
+++ b/dbtmetabase/parsers/dbt_manifest.py
@@ -29,7 +29,7 @@ class DbtManifestReader(DbtReader):
         schema = self.schema
         schema_excludes = self.schema_excludes
         includes = [each_string.upper() for each_string in self.includes]
-        excludes = self.excludes
+        excludes = [each_string.upper() for each_string in self.excludes]
 
         manifest = {}
 
--- a/dbtmetabase/parsers/dbt_manifest.py
+++ b/dbtmetabase/parsers/dbt_manifest.py
@@ -185,7 +185,9 @@ class DbtManifestReader(DbtReader):
 
         metabase_column: List[MetabaseColumn] = []
 
-        children = manifest["child_map"][model["unique_id"]]
+        unique_id = model["unique_id"]
+
+        children = manifest["child_map"][unique_id]
         relationship_tests = {}
 
         for child_id in children:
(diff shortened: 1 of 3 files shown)


dbtmetabase/parsers/dbt_manifest.py
```diff
@@ -28,7 +28,7 @@ def read_models(
         database = self.database
         schema = self.schema
         schema_excludes = self.schema_excludes
-        includes = self.includes
+        includes = [each_string.upper() for each_string in self.includes]
```
gouline, line 31: You shouldn't modify the list, do it at the point of comparison.

gouline (changes requested):
You explained about excludes but that leaves inconsistent behaviour so this pull request will only be accepted with both.

dbtmetabase/parsers/dbt_manifest.py
leoebfolsom: Thank you--I'd make the change right away, but I have a question about what that would look like.

I believe the point of comparison is `if (includes and model_name not in includes) or (model_name in excludes):`, which occurs twice in this file (lines 91 and 144).

Since `model_name` is always uppercase due to the transformation in line 49, wouldn't I have to transform the entire `includes` list to uppercase for `model_name not in includes` to evaluate properly?

I assume you don't want me to transform the list _twice_, so I'm not clear on what I should do differently, or why it's a problem to modify the list earlier on to avoid repetition since it is used later in the file twice.

I guess I could iterate through the list and transform each element as I go, but why write more code?

Not trying to be difficult here--I appreciate any clarification you can provide.

gouline: Ahhh, I see your point. In that case, yes, transform the whole list for includes and excludes. I would prefer if you lowercased them instead of uppercase (that's the preferred format in dbt).

Have a look at the CI error as well, I think the includes/excludes can be None so it warns that your list comprehension could crash.

leoebfolsom: Thank you! I think I've addressed everything you brought up. 

I also made one edit to `dbtmetabase/models/metabase.py` to accommodate special_type in MetabaseColumn. I think this was an accidental omission since special type is included in METABASE_META_FIELDS defined in line 9 of the same file. It would be awesome if that update could be included in this PR but I understand if not since it's unrelated.

gouline: Not accidental, Metabase renamed `special_type` to `semantic_type`. Please revert that change.

gouline:
Run `make check test` locally before pushing to make sure checks and tests are passing.

dbtmetabase/parsers/dbt_manifest.py
leoebfolsom: Ah, thank you for letting me know. I just reverted the change.

leoebfolsom:
@gouline I ran `make check test` and I believe everything passed, except for the following warning, which I don't think has to do with my code changes?
```
/dbt-metabase/tests/test_metabase.py:335: ResourceWarning: unclosed file <_io.TextIOWrapper name='tests/fixtures/lookups/table_lookups.json' mode='r' encoding='UTF-8'>
  baseline_table_lookups = json.load(
ResourceWarning: Enable tracemalloc to get the object allocation traceback
/dbt-metabase/tests/test_metabase.py:338: ResourceWarning: unclosed file <_io.TextIOWrapper name='tests/fixtures/lookups/field_lookups.json' mode='r' encoding='UTF-8'>
  baseline_field_lookups = json.load(
ResourceWarning: Enable tracemalloc to get the object allocation traceback
```

dbtmetabase/parsers/dbt_manifest.py
```diff
@@ -31,6 +31,11 @@ def read_models(
         includes = self.includes
         excludes = self.excludes
 
+        if includes is not None:
```
gouline, line 34: You don't need to assign it and then re-process, just replace the above assignments with 

```py
includes = [x.lower() for x in self.includes] if self.includes else []
excludes = [x.lower() for x in self.excludes] if self.excludes else []
```

leoebfolsom: That looks much better, thanks! Done.

gouline (approved):

PR #135, opened 2022-08-17
docs: Update the known table visibility_types

willbryant:


gouline:
I left them out on purpose, neither can be set in the UI, so not sure how supported they are.

willbryant:
They can now - it's these links on the right:
Screen Shot 2022-08-18 at 09 53 18

gouline (approved):

PR #139, opened 2022-08-22
feat: Support explicit foreign key relationships

gouline:
Tentatively support explicit foreign keys, without relationship tests:

```yaml
- name: user_id
  meta:
    metabase.foreign_key_to: ref('users')
    metabase.foreign_key_field: id
```

Works with **folder parser only** (for now, will consider adding to manifest parser in future).

### DIFF
--- a/dbtmetabase/parsers/dbt_folder.py
+++ b/dbtmetabase/parsers/dbt_folder.py
@@ -1,7 +1,7 @@
 import re
 import yaml
 from pathlib import Path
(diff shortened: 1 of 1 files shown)


willbryant:
I thought this is what the documented metabase.fk_target_table/metabase.fk_target_field thing already did?

gouline:
> I thought this is what the documented metabase.fk_target_table/metabase.fk_target_field thing already did?

That only works with manifest parser, not folder parser, but you make a good point that naming should be consistent. I'll change it to be the same.

PR #141, opened 2022-08-23
fix: Consistent model selection in parsing

### LINKED ISSUE #140: Schema option case sensitivity
Docs say you can use:

```
    dbt_schema_excludes:
      - development
      - testing
```

And there's also support for a `--dbt_schema_excludes` command-line option.

However, you must actually upcase the schema name, unlike the example before, because the code that actually looks at this is:

```
if schema_excludes and node["schema"].upper() in schema_excludes:
```

I'm not really clear why the `upper()` is there. If it's just for the sake of normalising, then we also need to normalise the arguments?

gouline:
Resolves #140

### DIFF
--- a/dbtmetabase/parsers/dbt_manifest.py
+++ b/dbtmetabase/parsers/dbt_manifest.py
@@ -169,8 +169,9 @@ class DbtManifestReader(DbtReader):
             dict -- One dbt model in Metabase-friendly format.
         """
 
-        metabase_column: List[MetabaseColumn] = []
+        metabase_columns: List[MetabaseColumn] = []
 
+        schema = model["schema"].upper()
         unique_id = model["unique_id"]
 
         children = manifest["child_map"][unique_id]
@@ -239,9 +240,10 @@ class DbtManifestReader(DbtReader):
                 }
 
         for _, column in model.get("columns", {}).items():
-            metabase_column.append(
+            metabase_columns.append(
                 self._read_column(
                     column=column,
+                    schema=schema,
                     relationship=relationship_tests.get(column["name"]),
                 )
             )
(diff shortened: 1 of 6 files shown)


gouline:
@willbryant See if this works for you before I merge and release

willbryant:
@gouline Nope it's actually doing completely NOOP runs on this branch. No further output after the "Session established successfully" banner.

Rolled back to 0.9.7 and it's working again, so don't think it's something wrong with my configuration.

gouline:
Drop any details of your configuration that I would need to reproduce and I'll look into it when I have time.

gouline:
Actually @willbryant give it another go, I fixed a silly typo that may have been causing the noop 🤔

willbryant:
It's still skipping for me. `-v` shows:

```
DEBUG    Skipping DIM_CUSTOMERS in schema BI not in target schema PUBLIC                                                                                                                                                                                    dbt_manifest.py:65
```

Which is not what the previous version did. It starts working again if I add:

```
dbt_schema: bi
```

To my config. Whereas before, I just had:

```
  dbt_schema_excludes:
  - bi_dbt_ga4
```

IOW I think the behavior change is that now effectively only one schema at a time is supported?

dbtmetabase/parsers/dbt_manifest.py
```diff
@@ -111,46 +97,41 @@ def read_models(
...
-                logger().debug("Skipping %s not of resource type source", model_name)
+            if source_database != self.database:
+                logger().debug(
+                    "Skipping %s not in target database %s", source_name, self.database
+                )
                 continue
 
-            if schema and node["schema"].upper() != schema.upper():
-                # Skip any models not in target schema
+            if self.schema and source_schema != self.schema:
```
willbryant, line 114: `self.schema` will always be set now I think?

gouline:
That makes sense, I was testing with the folder parser. Latest commit should fix the issue you raised.

willbryant:
Yup that's working - thanks.

PR #145, opened 2022-09-13
More column settings

willbryant:
Add support for the following properties:
* metabase.has_field_values
* metabase.coercion_strategy
* metabase.number_style (becomes settings -> number_style)


gouline:
Do they all have null defaults? Some Metabase API fields have non-null strings as defaults

willbryant:
No, they don't. Have I done that wrong? I'm only setting them if they are explicitly set on the DBT side.

gouline:
Just checking, looks good

gouline (approved):

PR #160, opened 2023-02-23
Improve Metabase display name handling

atvaccaro:
A couple small tweaks that we found useful in https://github.com/cal-itp/data-infra/tree/main/warehouse. Open to suggestions; would it be useful to only strip the whitespace if configured?

Fix: Replaces newlines in descriptions before syncing; we have some large model-level documentation blocks and Metabase just strips out the newlines between paragraphs, leaving no whitespace between paragraphs.

Feature: Parses `metabase.display_name` from the meta dictionary to allow for overriding; we use prefixes that do not look good in direct title-case, and overriding via the YAML makes it easy to change.

### DIFF
--- a/dbtmetabase/metabase.py
+++ b/dbtmetabase/metabase.py
@@ -355,7 +355,11 @@ class MetabaseClient:
 
         # Empty strings not accepted by Metabase
         model_display_name = model.display_name or None
-        model_description = model.description or None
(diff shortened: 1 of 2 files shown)


atvaccaro:
Oh I didn't realize this was already in the prior PR, sorry!

PR #161, opened 2023-03-10
feat: Add the 'cert' argument to the Requests framework

tomytrip:
To access the Metabase database in my company, we need to present a certificate to prove our identity. 
To do this, we need to use the Requests argument: "cert"
So I add it in the MetabaseClient object

### DIFF
--- a/dbtmetabase/metabase.py
+++ b/dbtmetabase/metabase.py
@@ -10,6 +10,7 @@ from typing import (
     Mapping,
     MutableMapping,
     Optional,
+    Tuple,
     Union,
 )
 
@@ -120,6 +121,7 @@ class MetabaseClient:
         user: str,
         password: str,
         verify: Optional[Union[str, bool]] = None,
+        cert: Optional[Union[str, Tuple[str, str]]] = None,
         session_id: Optional[str] = None,
         use_http: bool = False,
         sync: Optional[bool] = True,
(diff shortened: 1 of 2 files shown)


kozlek (approved):

dbtmetabase/metabase.py
```diff
@@ -144,11 +147,11 @@ def __init__(
         self.base_url = f"{'http' if use_http else 'https'}://{host}"
         self.session = requests.Session()
         self.session.verify = verify
+        self.session.cert = cert
         adaptor = HTTPAdapter(max_retries=Retry(total=3, backoff_factor=0.5))
         self.session.mount(self.base_url, adaptor)
         session_header = session_id or self.get_session_id(user, password)
         self.session.headers["X-Metabase-Session"] = session_header
-
```
gouline, line 151: Unrelated change, please revert

dbtmetabase/metabase.py
```diff
@@ -120,6 +121,7 @@ def __init__(
         user: str,
         password: str,
         verify: Optional[Union[str, bool]] = None,
+        cert: Optional[Union[str, Tuple[str, str]]] = None,
```
gouline, line 124: So it's either a string or a tuple of two strings?

tomytrip: Yes, you can verify it in the requests documentation at the Client Side Certificates part https://requests.readthedocs.io/en/latest/user/advanced/

gouline (approved):

PR #162, opened 2023-03-15
feat: add support for adding session_id without user and password

ryosagisu:
My company will disable password authentication and move every API call to use JWT authentication.
We already developed logic to get `session_id` value from JWT login.

These changes will remove `metabase_user` and `metabase_password` as required arguments. 
And adding validation to check if user provided either of `session_id` or both `metabase_user` and `metabase_password`.


dbtmetabase/models/exceptions.py
```diff
@@ -6,6 +6,10 @@ class NoDbtSchemaSupplied(Exception):
     """Thrown when using folder parser without supplying a schema"""
 
 
+class NoMetabaseCredentialsOrSessionIdSupplied(Exception):
```
gouline, line 9: Session ID is a type of credential too, so should just be `NoMetabaseCredentialsSupplied`

ryosagisu: updated

dbtmetabase/models/interface.py
```diff
@@ -84,6 +89,11 @@ def prepare_metabase_client(self, dbt_models: Optional[List[MetabaseModel]] = No
         if dbt_models is None:
             dbt_models = []
 
+        if not (self.user and self.password) and not self.session_id:
+            raise NoMetabaseCredentialsSupplied(
+                f"Credentials or Session ID is not supplied"
```
gouline, line 94:
```suggestion
                f"Credentials or session ID not supplied"
```

dbtmetabase/models/interface.py
```diff
@@ -84,6 +89,11 @@ def prepare_metabase_client(self, dbt_models: Optional[List[MetabaseModel]] = No
         if dbt_models is None:
             dbt_models = []
 
+        if not (self.user and self.password) and not self.session_id:
+            raise NoMetabaseCredentialsSupplied(
+                f"Credentials or session ID not supplied"
```
gouline, line 94:
```suggestion
                "Credentials or session ID not supplied"
```

dbtmetabase/models/interface.py
```diff
@@ -19,8 +23,8 @@ def __init__(
         self,
         database: str,
         host: str,
-        user: str,
-        password: str,
+        user: Optional[str] = None,
+        password: Optional[str] = None,
```
gouline, line 27: These type changes must be propagated down to `MetabaseClient`, that's why your CI is failing

gouline (changes requested):

dbtmetabase/models/interface.py
ryosagisu: updated, should passed the CI now

gouline (approved):

Issue #163, opened 2023-03-20
Shouldn't we sync the documentation even if tables schemas have mismatches between dbt manifest and metabase?


moreaupascal56:
Hi @gouline,

First thanks for the package ist has been really helpful, however the new releases 0.9.14 causes an error on our dbt project. 

```
    "Unable to align models between dbt target models and Metabase"
dbtmetabase.models.exceptions.MetabaseUnableToSync: Unable to align models between dbt target models and Metabase

```

I guess this means that a dbt model is not present in the metabase schema, but how can I identify which model is causin this ? 
Thanks ,
Have a great day

moreaupascal56:
Alright so I've looked more in details at the code and I understood what was going on, I had some warnings on non-existing columns like this message : 
`WARNING  Column colname not found in                          metabase.py:282
         SCHEMA.TABLE model
 `

Which cause `are_model_compatible = False` [here](https://github.com/gouline/dbt-metabase/blob/9c931434e75d48fd1933e680b95faa6ecf9827cf/dbtmetabase/metabase.py#L280), and then it stops the computation in MetabaseClient because `sync=True` per default. 

What I do not understand about this behaviour is that even for only one column mismatch the sync will not be run at all even if the majority of the tables schema matches. Maybe I just do not see the obvious reason ahha but why should we have the default   `sync=True` instead of  `sync=False` (as it was before) ? 
Thanks and have a great day :) 

PS: btw I (Pycharm actually 😅) saw a micro typo I'll open a really small MR to clean it.

gouline:
Intended behaviour is maintaining all columns in the models, hence sync is enabled by default (it was mistakenly turned off a few builds ago). You can run it in "best efforts" mode to export what it can by disabling it, but that's not the encouraged.

moreaupascal56:
Ok thanks that's more clear,  have a great day :)

PR #164, opened 2023-03-22
Remove unused statement in metabase.py

moreaupascal56:
Hey, 

As discussed here #163 just a small cleaning by Pycharm. 
As `sync_successful` will always be changed line 232 `sync_successful = self.models_compatible(models)`  because the first iteration of the loop will always be executed the assignment to False is unused.


gouline:
Not a typo, it indicates the default state is. Helpful if the loop becomes dependent on a condition.

moreaupascal56:
Alright yes I was wondering if it was an hint maybe.

Issue #170, opened 2023-06-21
Propagate dbt analyses to Metabase


keurcien:
Thanks for the great package, it gives a lot of ideas!

One of the questions I have, is whether you have considered to propagate dbt analyses as Metabase cards?
It would be great to be able to have a collection of cards that have been versioned with dbt. 

I believe the general idea would be to have a 1-to-1 mapping between an analysis and a Metabase card, and a solution could look like this:

- parse the `manifest.json` to look for exposures.
- have a way to identify that an exposure is a Metabase card (maybe in the way the exposure name is defined, e.g add a prefix to parse, `- name: MB117 - ....` for the 117th card in Metabase)
- the `depends_on` should be a ref to a unique analysis
- get the analysis SQL and update the card `dataset_query` property.

N.B: I know analyses aren't refable: [issue](https://github.com/dbt-labs/dbt-core/issues/7127), but the idea remains the same using ephemeral models that could be stored in a `metabase` or `metabase_cards` subfolder. If this looks like an interesting feature, I'd love to contribute if needed.

gouline:
Sounds interesting, would love to see a draft of this implementation.

PR #171, opened 2023-06-23
feat: propagate dbt models as Metabase cards

keurcien:
This PR is a quick draft of what I had in mind when creating this issue: https://github.com/gouline/dbt-metabase/issues/170#issuecomment-1601967111

The main drawbacks I see here:
- syncable exposures should follow a specific naming convention "MB...", maybe a meta field would be more appropriate
- can only update cards, no creation


gouline:
Reminder that you still have CI failures that you need to fix if you want this to be reviewed. Otherwise, feel free to close.

Issue #168, opened 2023-05-29
use human friendly exposure names


geoHeil:
Starting in v1.3, the 'name' of an exposure should contain only letters,
numbers, and underscores. Exposures support a new property, 'label', which may
contain spaces, capital letters, and special characters.
FOO_BAR,_baz,_some_stuff_foo does not follow this
pattern. Please update the 'name', and use the 'label' property for a
human-friendly title. This will raise an error in a future version of dbt-core.

PR #172, opened 2023-08-01
fixed exposure names and added labels

mcnick:
# Problem
https://github.com/gouline/dbt-metabase/issues/168 Since dbt 1.3 exposure names containing other characters than letters, numbers, underscores and hyphens are deprecated

# Changes
- Replaced all characters not being letters, numbers or hyphens by underscores in the exposure name
- Added the original card name as exposure label

### DIFF
--- a/dbtmetabase/metabase.py
+++ b/dbtmetabase/metabase.py
@@ -731,7 +731,7 @@ class MetabaseClient:
 
                 exposure_label = exposure_name
                 # Only letters, numbers, underscores and hyphens allowed in model names in dbt docs DAG / No duplicate model names
-                exposure_name = re.sub("[^\w-]", "_", exposure_name)
+                exposure_name = re.sub(r"[^\w-]", "_", exposure_name)
                 enumer = 1
                 while exposure_name in documented_exposure_names:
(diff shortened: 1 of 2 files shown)


dbtmetabase/metabase.py
```diff
@@ -729,8 +729,9 @@ def increase_indent(self, flow=False, indentless=False):
                     creator_email = creator.get("email")
                     creator_name = creator.get("common_name")
 
-                # No spaces allowed in model names in dbt docs DAG / No duplicate model names
-                exposure_name = exposure_name.replace(" ", "_")
+                exposure_label = exposure_name
+                # Only lower case letters and underscores allowed in model names in dbt docs DAG / No duplicate model names
+                exposure_name = re.sub("[^a-z0-9]", "_", exposure_name.lower())
```
gouline, line 734: How does this affect non-Latin characters? Are they definitely not allowed in dbt?

mcnick: Good point, they would be replaced.
I also checked the dbt code again, the warning message is actually slightly wrong/misleading. So I adapted the replacement to be equivalent to the check they're doing.

gouline (approved):

PR #174, opened 2023-08-31
Improvements

elongl:
1. The [first change](https://github.com/gouline/dbt-metabase/commit/2dbc756da67b9e64cc78070f81b04ec1b3d4a3a1) also searches for dbt property files in `models/*.yaml` and not only `models/*.yml`.
2. The [second change](https://github.com/gouline/dbt-metabase/commit/6b5246136a3a8cd22f443485bbb6c6476181e8de) corrects the format of exposures.
2.1 In dbt 1.3.0 onwards, you're not allowed to have upper-case characters in the exposure's name.
2.2 If a dashboard has two entities that query the same table, they'll be a duplicate in `depends_on`. I applied a dedup for that.
3. Listing dbt tables (models, sources) using `dbt ls`.
The main motivation for doing so is being able to safely get dbt models that are not defined in the property YAML files.
An alternative would've been to parse `models/*.sql` files but that's difficult due to the `alias` config that can override the table name from the default file name.
Using `dbt ls` allows us to safely get the table name.

### DIFF
--- a/dbtmetabase/metabase.py
+++ b/dbtmetabase/metabase.py
@@ -727,7 +727,7 @@ class MetabaseClient:
 
                 exposure_label = exposure_name
                 # Only letters, numbers, underscores and hyphens allowed in model names in dbt docs DAG / No duplicate model names
-                exposure_name = re.sub(r"[^\w-]", "_", exposure_name)
+                exposure_name = re.sub(r"[^\w-]", "_", exposure_name).lower()
                 enumer = 1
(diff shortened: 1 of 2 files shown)


gouline:
Can you please split these into separate pull requests? 

They're not related to each other, so if there are issues attributed to them down the line, we need to be able to roll back to specific changes.

elongl:
For sure, definitely makes more sense. Closing.

PR #175, opened 2023-09-01
Exposure updated format

elongl:


dbtmetabase/metabase.py
```diff
@@ -951,11 +945,15 @@ def _build_exposure(
                 "name": creator_name,
                 "email": creator_email or "",
             },
-            "depends_on": [
-                refable_models[exposure.upper()]
-                for exposure in list({m for m in self.models_exposed})
-                if exposure.upper() in refable_models
-            ],
+            "depends_on": list(
```
elongl, line 948: Deduping in case a dashboard has two views on the same table.

dbtmetabase/metabase.py
```diff
@@ -645,7 +645,6 @@ def increase_indent(self, flow=False, indentless=False):
         parsed_exposures = []
 
         for collection in self.collections:
-
```
gouline, line 648: Revert irrelevant changes.

gouline (changes requested):
Revert all irrelevant changes, provide a description for the PR explaining what you're doing any why, and run checks/tests locally before pushing, to speed up the process, since I have to manually approve CI runs for new contributors.

PR #176, opened 2023-09-01
Listing dbt tables (models, sources) using `dbt ls`.

elongl:
The main motivation for doing so is being able to safely get dbt models that are not defined in the property YAML files. An alternative would've been to parse `models/*.sql` files but that's difficult due to the `alias` config that can override the table name from the default file name. Using `dbt ls` allows us to safely get the table name.


gouline:
Folder parser is the legacy implementation that doesn't require running dbt, it's left as-is for backwards compatibility. For more precise results, use the manifest parser instead.

PR #187, opened 2023-11-22
feat: configurable metabase api timeotus

### LINKED ISSUE #179: Configure API timeout for calls made to Metabase
Currently all calls are made to the Metabase API using the function api() defined on line 961 in dbtmetabase/metabase.py.

When any API is called below is the code executed
`response = self.session.request(
            method,
            f"{self.base_url}{path}",
            timeout=15,
            **kwargs,
        )`
        
So for eg. when the database/{id}/metadata?include_hidden=True is executed there is a timeout for only 15 seconds.

We have a lot of tables in our database and this operation alone is taking almost 2 minutes to complete.

We always face this 15 seconds timeout and are not able to proceed and update any metadata on Metabase from DBT.

Will this require a code fix or is there a workaround I could use.

Link: https://github.com/gouline/dbt-metabase/blob/master/dbtmetabase/metabase.py#L961

pquadri:
Closes #179

### DIFF
--- a/dbtmetabase/__init__.py
+++ b/dbtmetabase/__init__.py
@@ -24,6 +24,7 @@ ENV_VARS = [
     "MB_HOST",
     "MB_DATABASE",
     "MB_SESSION_TOKEN",
+    "MB_HTTP_TIMEOUT",
 ]
 
 
@@ -288,6 +289,15 @@ def shared_opts(func: Callable) -> Callable:
         type=(str, str),
         multiple=True,
         help="Additional HTTP request header to be sent to Metabase.",
+    )
+    @click.option(
+        "--metabase_http_timeout",
+        cls=OptionAcceptableFromConfig,
+        type=int,
+        default=15,
+        envvar="MB_HTTP_TIMEOUT",
+        show_envvar=True,
+        help="Set the value for single requests timeout",
     )
     @functools.wraps(func)
     def wrapper(*args, **kwargs):
@@ -559,6 +569,7 @@ def models(
     metabase_sync: bool = True,
     metabase_sync_timeout: Optional[int] = None,
     metabase_exclude_sources: bool = False,
+    metabase_http_timeout: int = 15,
     dbt_include_tags: bool = True,
     dbt_docs_url: Optional[str] = None,
     verbose: bool = False,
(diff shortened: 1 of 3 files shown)


dbtmetabase/metabase.py
```diff
@@ -723,7 +724,10 @@ def increase_indent(self, flow=False, indentless=False):
                         creator = self.api("get", f"/api/user/{exposure['creator_id']}")
                     except requests.exceptions.HTTPError as error:
                         creator = {}
-                        if error.response.status_code != 404:
+                        if (
+                            error.response is not None
+                            and error.response.status_code != 404
+                        ):
```
pquadri, line 730: this was updated due to mypy complaining

dbtmetabase/__init__.py
```diff
@@ -1,14 +1,14 @@
...
 import os
+from pathlib import Path
+from typing import Any, Callable, Dict, Iterable, Optional
 
 import click
 import yaml
 
 from .logger import logging as package_logger
-from .models.interface import MetabaseInterface, DbtInterface
+from .models.interface import DbtInterface, MetabaseInterface
```
gouline, line 11: Not relevant, please revert

dbtmetabase/__init__.py
```diff
@@ -289,6 +311,13 @@ def shared_opts(func: Callable) -> Callable:
         multiple=True,
         help="Additional HTTP request header to be sent to Metabase.",
     )
+    @click.option(
+        "--metabase_requests_timeout",
+        cls=OptionAcceptableFromConfig,
+        type=INT_OR_NONE,
```
gouline, line 317: There should never be a none value. If you personally don't need one, you can set it to say 24 hours, but there should always be some value.

dbtmetabase/__init__.py
```diff
@@ -685,6 +717,7 @@ def exposures(
     metabase_verify: Optional[str] = None,
     metabase_sync: bool = True,
     metabase_sync_timeout: Optional[int] = None,
+    metabase_requests_timeout: Optional[int] = 15,
```
gouline, line 720: As above, shouldn't be `Optional` (keep the default value though, obviously).

dbtmetabase/__init__.py
```diff
@@ -289,6 +311,13 @@ def shared_opts(func: Callable) -> Callable:
         multiple=True,
         help="Additional HTTP request header to be sent to Metabase.",
     )
+    @click.option(
+        "--metabase_requests_timeout",
```
gouline, line 315: I would prefer `metabase_http_timeout` for the naming.

dbtmetabase/metabase.py
gouline: We probably want to raise if it's none too, so:

```py
if not error.response or error.response.status_code != 404:
  raise
```

gouline (changes requested):

dbtmetabase/__init__.py
pquadri: i have no idea why i read in the issue that it could be None, fixing it

gouline (approved):
Very clean, thanks!

bulatyapparov:
How to use this?

PR #188, opened 2023-12-01
Allow custom http adapter for certificate handling

alexisflipo:
First of all, thank you for the development of this library. We use it often in the company i work for. 

### RATIONALE

Since we changed the authentication from Airflow to Metabase we reached some issues even with the latest version. This PR aims to add an other way to handle certificates but without needing them to be physically present (since the path is required on cert parameter). I used the `requests-pkcs12` library to handle this behavior

### CONSTRAINTS

- The certificate is not present on the machine
- We can only authenticate with the certificate by its content and its password

### SOLUTION

- Add the `http_adapter` parameter in MetabaseInterface and client as optional
- Allows to provide a custom http_adapter 

Thanks,

Don't hesitate to contact me if you look for further informations

### DIFF
--- a/dbtmetabase/metabase.py
+++ b/dbtmetabase/metabase.py
@@ -28,8 +28,6 @@ from .models.metabase import (
     NullValue,
 )
 
-from requests_pkcs12 import Pkcs12Adapter
-
 
 class MetabaseClient:
     """Metabase API client."""
@@ -124,14 +122,14 @@ class MetabaseClient:
         password: Optional[str],
         verify: Optional[Union[str, bool]] = None,
         cert: Optional[Union[str, Tuple[str, str]]] = None,
-        pkcs12_data: Optional[Tuple[str, str]] = None,
+        pkcs12_data: Optional[Union[str, Tuple[str, str]]] = None,
         session_id: Optional[str] = None,
         use_http: bool = False,
         sync: Optional[bool] = True,
         sync_timeout: Optional[int] = None,
         exclude_sources: bool = False,
         http_extra_headers: Optional[dict] = None,
-        http_timeout: Optional[int] = 15,
+        http_timeout: int = 15,
     ):
         """Constructor.
 
@@ -153,6 +151,7 @@ class MetabaseClient:
         """
 
         self.base_url = f"{'http' if use_http else 'https'}://{host}"
+        self.http_timeout = http_timeout
         self.session = requests.Session()
         self.session.verify = verify
         self.session.cert = cert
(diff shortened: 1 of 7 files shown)


gouline (changes requested):
I'm not a fan of adding a non-major dependency for a very limited use case. My preference would be to implement this as an extra (i.e. `pip install dbt-metabase[pkcs12]`).

Alternatively, you could allow for a custom `HTTPAdapter` to be provided, since it looks like your whole implementation ends up in swapping it out for `Pkcs12Adapter`.

dbtmetabase/metabase.py
```diff
@@ -122,13 +124,14 @@ def __init__(
...
         cert: Optional[Union[str, Tuple[str, str]]] = None,
+        pkcs12_data: Optional[Tuple[str, str]] = None,
         session_id: Optional[str] = None,
         use_http: bool = False,
         sync: Optional[bool] = True,
         sync_timeout: Optional[int] = None,
         exclude_sources: bool = False,
         http_extra_headers: Optional[dict] = None,
-        http_timeout: int = 15,
+        http_timeout: Optional[int] = 15,
```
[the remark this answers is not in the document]
alexisflipo: Thanks, changed !

requirements.txt
```diff
@@ -2,3 +2,4 @@ PyYAML>=5.4.1
 requests>=2.26.0
 click>=8.0.0
 rich>=12.0.0
+requests-pkcs12==1.22
```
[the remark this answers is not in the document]
alexisflipo: The constraints are changed, thanks !

dbtmetabase/metabase.py
```diff
@@ -28,6 +28,8 @@
     NullValue,
 )
 
+from requests_pkcs12 import Pkcs12Adapter
```
gouline, line 31: This will fail for those who don't install the extra

gouline (commented):
Honestly, given the whole fuss with extras and making sure nothing breaks, I'm leaning towards just allowing for a custom HTTP adapter, as I mentioned above.

Then no need for a new dependency, you can create the PKCS adapter in your own code and pass it through the config. Other advanced users could have use for that to set something else custom.

gouline:
CI checks also still failing, make sure they pass locally before you push.

alexisflipo:
Yep got it. I changed it to make the http_adapter optional. I added the `requests` import in the `inteface.py` to add the right class in the type hint.

dbtmetabase/models/interface.py
```diff
@@ -1,17 +1,18 @@
...
     NoDbtPathSupplied,
     NoDbtSchemaSupplied,
     NoMetabaseCredentialsSupplied,
 )
-from ..parsers.dbt import DbtReader
-from ..parsers.dbt_folder import DbtFolderReader
-from ..parsers.dbt_manifest import DbtManifestReader
-from ..metabase import MetabaseClient
+from .metabase import MetabaseModel
+from requests.adapters import BaseAdapter
```
gouline, line 15: What are you using for import ordering? I might need to add this to the checks, to make sure it's consistent.

dbtmetabase/metabase.py
```diff
@@ -129,6 +129,7 @@ def __init__(
         exclude_sources: bool = False,
         http_extra_headers: Optional[dict] = None,
         http_timeout: int = 15,
+        http_adapter: Optional[BaseAdapter] = None,
```
gouline, line 132: You don't need BaseAdapter, just use HTTPAdapter. That's what your PKCS library extends as well.

gouline (commented):
Almost there. Please update the title/description of the PR to reflect the actual changes: you can leave the justification for PKCS, but describe the implementation as it is now.

dbtmetabase/models/interface.py
alexisflipo: Ok nice ! I'm using `ruff`

dbtmetabase/metabase.py
alexisflipo: Thanks

gouline (approved):

PR #194, opened 2023-12-08
Refactor CLI setup

gouline:
- Refactor Click CLI setup, including the following breaking changes:
  - Options are now separated with hyphens, not underscores (e.g. `--dbt_database` becomes `--dbt-database`)
  - Lists are now separated with commas, not spaces (e.g. `--dbt-includes model1 model2` becomes `--dbt-includes model1,model2`)
  - Metabase environment variables now prefixed with `METABASE_` instead of `MB_`
  - Interactive config generation has been removed (used to be invoked via `dbt-metabase config`)
  - Miscellaneous options were renamed for clarity (e.g. `--dbt-path` is now `--dbt-project-path`)
- Support for custom config file, in addition to the default location ~/.dbt-metabase/config.yml in the same format (e.g. `dbt-metabase --config-file path/to/config.yml models`)
- Migrate from `script` to `entry_points` in setup.py for better cross-platform support


dbtmetabase/__init__.py
```diff
@@ -1,785 +1,12 @@
...
-        type=int,
-        help="Synchronization timeout (in secs). If set, we will fail hard on synchronization failure; if not set, we will proceed after attempting sync regardless of success. Only valid if sync is enabled",
-    )
-    @click.option(
-        "--http_extra_headers",
-        cls=OptionAcceptableFromConfig,
-        type=(str, str),
-        multiple=True,
-        help="Additional HTTP request header to be sent to Metabase.",
-    )
```
tanguyantoine, line 292: Hi, I see that this part has been removed from the new version 1.0.0, is it expected?

gouline: Still available in the programmatic API. I couldn't get the dictionary type to work via CLI flag, were you using it via config YAML file?

gouline: Released in v1.0.1, you can use `--http-header x-some-header some-value` or via YAML config:

```yaml
config:
  http_headers:
    - [x-some-header, some-value]
```

tanguyantoine:
That was fast, thank you 🙏

PR #226, opened 2024-02-14
fix exposures crashing with referencing questions with multiple variables

kkuznets:
**Problem**
Running `dbt-metabase exposures` produces faulty `exposures.json` if the referenced Metabase questions contain multiple variables in one query.

**To reproduce**
The problem is due to the greedy nature of the `.*` regex pattern that will catch multiple sets of double braces if they exist within the same line.

_Example query:_
```
SELECT *
FROM generate_series(  DATE_TRUNC ('day',{{start_date}}  - cast({{rolling_days}} || ' days' as interval)),  DATE_TRUNC ('day',now() - interval '1 day'), '1 day') as buckets(day);
```

**Solution**
Make the regex less greedy to match as little text as possible.

### DIFF
--- a/dbtmetabase/format.py
+++ b/dbtmetabase/format.py
@@ -142,4 +142,4 @@ def safe_description(text: Optional[str]) -> str:
     Returns:
         str: Sanitized string with escaped Jinja syntax.
     """
-    return re.sub(r"{{(.*)}}", r"(\1)", text or "")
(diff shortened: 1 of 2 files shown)


gouline:
Nice! Can you please add a unit test that fails before your fix and succeeds after?

kkuznets:
Sure! Added a simple unit test.

gouline (approved):

PR #238, opened 2024-02-28
Skip missing cards/dashboards in exposures

### LINKED ISSUE #237: Exposure Extraction Error | In Case of Missing Cards in collection
**Context:** 

When I use Metabase API to delete archived cards, it creates inconsistency because card are still referenced in collections.
So [at this step](https://github.com/gouline/dbt-metabase/blob/9fc8bcd67e6f418bb45561c238c8ffa6dd9dfb49/dbtmetabase/_exposures.py#L98C6-L98C68), it's possible to search a deleted card/dashboard causing an HTTP code 400 error and makes application crash.

**Problem**

In case of inconsistency between collections and cards/dashboards, dbt-metabase will crash because it tries to query a missing card/dashboard. 

**Request**

Skip a card/dashboard which is referenced in a collection but has been deleted.

**Technical Solution**

Test for each items in a Collection if the associated resource is reachable otherwise continue to the next item.
We could add a try&catch statement for get_dashboard and get_card as we did [here](https://github.com/gouline/dbt-metabase/blob/9fc8bcd67e6f418bb45561c238c8ffa6dd9dfb49/dbtmetabase/metabase.py#L166)

**Note**

Event if it's not the purpose of that tool, it could be interresting to highlight inconsistency with the logger in case of unreachable item resource.

dosnep:
- Fix: dbt-metabase now skips unreachable resources {card, dashboard, user} in case of **requests.exceptions.HTTPError: 404 Client Error: Not Found for url**

Closes #237

### DIFF
--- a/dbtmetabase/metabase.py
+++ b/dbtmetabase/metabase.py
@@ -145,7 +145,13 @@ class Metabase:
 
     def get_card(self, uid: str) -> Mapping:
         """Retrieves card (known as question in Metabase UI)."""
-        return dict(self._api("get", f"/api/card/{uid}"))
+        try:
+            return dict(self._api("get", f"/api/card/{uid}"))
+        except requests.exceptions.HTTPError as error:
+            if error.response.status_code == 404:
+                _logger.warning("Card '%s' not found", uid)
+                return None
+            raise
 
     def format_card_url(self, uid: str) -> str:
         """Formats URL link to a card (known as question in Metabase UI)."""
@@ -153,7 +159,13 @@ class Metabase:
 
     def get_dashboard(self, uid: str) -> Mapping:
         """Retrieves dashboard."""
-        return dict(self._api("get", f"/api/dashboard/{uid}"))
+        try:
+            return dict(self._api("get", f"/api/dashboard/{uid}"))
+        except requests.exceptions.HTTPError as error:
+            if error.response.status_code == 404:
+                _logger.warning("Dashboard '%s' not found", uid)
+                return None
+            raise
 
     def format_dashboard_url(self, uid: str) -> str:
         """Formats URL link to a dashboard."""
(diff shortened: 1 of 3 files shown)


dbtmetabase/metabase.py
```diff
@@ -145,15 +145,27 @@ def get_collection_items(
 
     def get_card(self, uid: str) -> Mapping:
```
gouline, line 146:
```suggestion
    def find_card(self, uid: str) -> Optional[Mapping]:
```

dbtmetabase/metabase.py
```diff
@@ -145,15 +145,27 @@ def get_collection_items(
...
+            if error.response.status_code == 404:
+                _logger.warning("Card '%s' not found", uid)
+                return None
+            raise
 
     def format_card_url(self, uid: str) -> str:
         """Formats URL link to a card (known as question in Metabase UI)."""
         return f"{self.url}/card/{uid}"
 
     def get_dashboard(self, uid: str) -> Mapping:
```
gouline, line 160:
```suggestion
    def find_dashboard(self, uid: str) -> Optional[Mapping]:
```

gouline (changes requested):
Please read [CONTRIBUTING.md](https://github.com/gouline/dbt-metabase/blob/master/CONTRIBUTING.md#pull-requests) for local checks to run before committing, to avoid waiting for failed CI.

dosnep:
> ommitting, to avoid waiting for failed CI.

CI has been tested locally (fmt-check and unit tests)

gouline:
Please read [CONTRIBUTING.md](https://github.com/gouline/dbt-metabase/blob/master/CONTRIBUTING.md#pull-requests) **carefully**. You only ran the formatting check, you have to run all of them.

PR #242, opened 2024-03-13
Handle uppercase references from Metabase GUI derived queries

alex-float-on:
Relates to discussion: https://github.com/gouline/dbt-metabase/discussions/241

Some data warehouses, such as snowflake, store their object names (database, schema, table, columns) in uppercase ([ref](https://community.snowflake.com/s/article/faq-when-i-retrieve-database-schema-table-or-column-names-why-does-snowflake-display-them-in-uppercase)). 

This means that metabase questions derived from the visual editor (GUI) need to have their sources lowercased in order to match with dbt's models and nodes. We already do this for SQL native queries (refs: [1](https://github.com/gouline/dbt-metabase/blob/master/dbtmetabase/_exposures.py#L250), [2](https://github.com/gouline/dbt-metabase/blob/master/dbtmetabase/_exposures.py#L255)) so this is just a continuation of that logic.

## Update

If `table.json` fixture is modified to contain uppercase table names, the exposure tests fail before this change is applied and pass after it is applied.

### DIFF
--- a/dbtmetabase/_exposures.py
+++ b/dbtmetabase/_exposures.py
@@ -214,7 +214,9 @@ class ExposuresMixin(metaclass=ABCMeta):
                     )
                 elif query_source in ctx.table_names:
                     # Normal question
-                    source_table = ctx.table_names.get(query_source).lower()
+                    source_table = ctx.table_names.get(query_source)
+                    if source_table:
+                        source_table = source_table.lower()
                     _logger.info("Extracted model '%s' from card", source_table)
                     depends.add(source_table)
 
--- a/dbtmetabase/_exposures.py
+++ b/dbtmetabase/_exposures.py
@@ -214,7 +214,7 @@ class ExposuresMixin(metaclass=ABCMeta):
                     )
                 elif query_source in ctx.table_names:
                     # Normal question
-                    source_table = ctx.table_names.get(query_source)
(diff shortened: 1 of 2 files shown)


dbtmetabase/_exposures.py
```diff
@@ -214,7 +214,7 @@ def __extract_card_exposures(
                     )
                 elif query_source in ctx.table_names:
                     # Normal question
-                    source_table = ctx.table_names.get(query_source)
+                    source_table = ctx.table_names.get(query_source).lower()
```
gouline, line 217: It could be None, add a similar `if source_table:` check that you saw with `joined_table` below

gouline (changes requested):
Read [CONTRIBUTING.md](https://github.com/gouline/dbt-metabase/blob/master/CONTRIBUTING.md#pull-requests) carefully and run all the checks locally before committing code.

gouline (approved):

gouline:
> One thing I noticed, however, is that we don't actively test table join exposures coming from non-sql questions. The only metabase question with a join is [card 23](https://github.com/gouline/dbt-metabase/blob/master/tests/fixtures/api/card/23.json#L449) but this card also happens to be the only one that is missing from the [expected exposures](https://github.com/gouline/dbt-metabase/tree/master/tests/fixtures/exposure/type/card).

I fixed this in #244.

PR #243, opened 2024-03-13
Add flag to exposures for excluding all unverified collection items

alex-float-on:
This PR introduces the ability to add an `--exclude-unverified-cards` flag to the `exposures` command in order to weed out exposures from cards that were not verified through metabase.

Note: I tried but could not figure out how to implement a unit test for this.

### DIFF
--- a/dbtmetabase/__main__.py
+++ b/dbtmetabase/__main__.py
@@ -360,11 +360,11 @@ def models(
     help="Include personal Metabase collections.",
 )
 @click.option(
-    "--exclude-unverified-cards",
-    envvar="EXCLUDE_UNVERIFIED_CARDS",
+    "--exclude-unverified",
+    envvar="EXCLUDE_UNVERIFIED",
     show_envvar=True,
     is_flag=True,
-    help="Exclude cards that have not been verified.",
+    help="Exclude items that have not been verified. Only applies to entity types that support verification.",
 )
 def exposures(
     output_path: str,
@@ -372,7 +372,7 @@ def exposures(
     include_collections: Optional[Sequence[str]],
     exclude_collections: Optional[Sequence[str]],
     allow_personal_collections: bool,
-    exclude_unverified_cards: bool,
+    exclude_unverified: bool,
     core: DbtMetabase,
 ):
     core.extract_exposures(
@@ -383,7 +383,7 @@ def exposures(
             exclude=exclude_collections,
         ),
         allow_personal_collections=allow_personal_collections,
-        exclude_unverified_cards=exclude_unverified_cards,
+        exclude_unverified=exclude_unverified,
     )
 
 
--- a/dbtmetabase/__main__.py
+++ b/dbtmetabase/__main__.py
@@ -360,11 +360,11 @@ def models(
     help="Include personal Metabase collections.",
 )
 @click.option(
-    "--exclude-unverified",
-    envvar="EXCLUDE_UNVERIFIED",
+    "--exclude-unverified-cards",
+    envvar="EXCLUDE_UNVERIFIED_CARDS",
     show_envvar=True,
     is_flag=True,
-    help="Exclude collection items that have not been verified.",
+    help="Exclude cards that have not been verified.",
 )
 def exposures(
     output_path: str,
@@ -372,7 +372,7 @@ def exposures(
     include_collections: Optional[Sequence[str]],
     exclude_collections: Optional[Sequence[str]],
     allow_personal_collections: bool,
-    exclude_unverified: bool,
+    exclude_unverified_cards: bool,
     core: DbtMetabase,
 ):
     core.extract_exposures(
@@ -383,7 +383,7 @@ def exposures(
             exclude=exclude_collections,
         ),
         allow_personal_collections=allow_personal_collections,
-        exclude_unverified=exclude_unverified,
+        exclude_unverified_cards=exclude_unverified_cards,
     )
 
 
--- a/dbtmetabase/__main__.py
+++ b/dbtmetabase/__main__.py
@@ -359,12 +359,20 @@ def models(
     is_flag=True,
     help="Include personal Metabase collections.",
 )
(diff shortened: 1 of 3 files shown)


dbtmetabase/metabase.py
```diff
@@ -131,6 +131,7 @@ def get_collection_items(
         self,
         uid: str,
         models: Sequence[str],
+        exclude_unverified: bool = False,
```
gouline, line 134: This object is strictly a wrapper around Metabase API, don't add extra business logic (especially command-specific business logic) to it. You filter it at the point of invocation.

dbtmetabase/metabase.py
```diff
@@ -140,6 +141,16 @@ def get_collection_items(
...
+        if exclude_unverified:
+            verified_results = list(
+                self._api(
+                    method="get",
+                    path=f"/api/search/",
+                    params={"models": models, "verified": True},
+                )
+            )
+            verified_ids = [y['id'] for y in verified_results]
+            results = [x for x in results if x['id'] in verified_ids]
```
gouline, line 153: Unnecessary and wasteful, verified cards have a field `moderated_status` set to `"verified"`.

What happens to dashboards? They can't be verified, so will this just exclude all dashboards?

gouline (changes requested):

dbtmetabase/__main__.py
```diff
@@ -359,12 +359,20 @@ def models(
     is_flag=True,
     help="Include personal Metabase collections.",
 )
+@click.option(
+    "--exclude-unverified-cards",
```
gouline, line 363: I think it should still be called `--exclude-unverified` in case dashboards ever become verified. "Cards" is also a term I avoid, because it doesn't appear anywhere in the UI (they're called questions), so Metabase users unfamiliar with its API won't know what it means

dbtmetabase/_exposures.py
```diff
@@ -89,6 +91,13 @@ def extract_exposures(
                 uid=collection["id"],
                 models=("card", "dashboard"),
             ):
+                if (
+                    exclude_unverified_cards
+                    and item.get("moderated_status") != "verified"
+                    and item.get("model") == "card"
```
gouline, line 97:
```suggestion
                    item["model"] == "card"
                    and exclude_unverified_cards
                    and item.get("moderated_status") != "verified"
```

The `model` key must be there, otherwise we have bigger problems. And operators are lazy, so in order of specificity.

dbtmetabase/_exposures.py
```diff
@@ -89,6 +91,13 @@ def extract_exposures(
...
+                ):
+                    _logger.debug("Skipping unverified card '%s'", item["name"])
+                    continue
```
gouline, line 100:
```suggestion
                    continue

```

gouline (changes requested):

dbtmetabase/__main__.py
alex-float-on: > should still be called `--exclude-unverified` in case dashboards ever become verified

I was internally debating this before pushing the updated change. The issue with keeping the exclusion general to all model types is that it creates ambiguity for model types that *can't* be verified. For example, is a dashboard considered unverified? Or were you implying we should actually change the logic to exclude dashboards altogether when this flag is enabled?

> "Cards" is also a term I avoid [...] Metabase users unfamiliar with its API won't know what it means.

That's very true; I agree 💯 . Given the point I made above, I think we should call the flag `--exclude-unverified-questions`. Thoughts?

dbtmetabase/_exposures.py
alex-float-on: Since this is a chain of `and` conditions, we should order the conditions from the most likely to be False to the least likely, right?

I figured the `exclude_unverified_cards` would by far be the least likely since it won't be used by most users. That's why I placed it first. I placed `and item.get("model") == "card"` last because I figured almost all items being iterated over are cards. Happy to defer to your expertise if you disagree though 👍

gouline: Good point about `exclude`, in that case:

```suggestion
                    exclude_unverified_cards
                    and item["model"] == "card"
                    and item.get("moderated_status") != "verified"
```

Otherwise, you're checking a model-specific value before verifying that it's a model that can have that value.

alex-float-on: Very true!

dbtmetabase/__main__.py
gouline: No, I'm happy with only applying filtering to questions for now.

What I want to avoid is adding too many niche settings (considering nobody had requested this feature before), e.g. if/when dashboards can be verified, instead of adding yet another setting `--exclude-unverified-dashboards`, I'd rather just announce a behaviour change that this now filters dashboards too.

alex-float-on: Given what you're saying, I'm starting to think we want an `--include-verified-only` flag instead and strictly filter on anything that has the verification status. This seems like the only option that gives us both 1) clarity in current functionality and 2) future-proof API.

alex-float-on: But again, very happy to keep the current logic and revert the flag's name to `--exclude-unverified` if that's what you prefer. We can make it clear in the documentation what this flag does.

gouline: "Exclude unverified" and "only include verified" are logically identical, I prefer the former because it's shorter. Neither stipulates how to handle unverifiable entities (e.g. dashboards), so we have to make a decision either way (i.e. we include unverifiable entities regardless).

alex-float-on: pushed the change back to `--exclude-unverified`. Let me know if you're good with the wording of the help docs.

gouline (approved):
Looks good, thanks for making those changes!

I'll merge this on Monday, I want to release your other fix as a patch release (1.1.2) and this change will go in the next minor release (1.2.0).

PR #246, opened 2024-03-27
Fix Dashboard Exposures Issue

yasminadimy:
Hello! 👋
I hope this message finds you well.

I recently encountered an issue in my environment where I wasn't able to retrieve the dashboard exposures, whereas the functionality worked seamlessly for cards. Upon investigating, I suspect that there might have been a change in the reference handling, which could be causing this discrepancy.

Here is a PR to handle the change and ensure consistent retrieval of both dashboard and card exposures!

Your feedback and review of the proposed changes would be greatly appreciated. Please feel free to reach out if you have any questions or suggestions regarding this PR ! ☀️

### DIFF
--- a/dbtmetabase/_exposures.py
+++ b/dbtmetabase/_exposures.py
@@ -126,7 +126,7 @@ class ExposuresMixin(metaclass=ABCMeta):
                         continue
 
                     entity = dashboard_entity
-                    cards = entity.get("dashcards", [])
+                    cards = entity.get("dashcards", entity.get("ordered_cards", []))
                     if not cards:
                         continue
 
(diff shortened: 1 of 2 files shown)


dbtmetabase/_exposures.py
```diff
@@ -126,7 +126,7 @@ def extract_exposures(
                         continue
 
                     entity = dashboard_entity
-                    cards = entity.get("ordered_cards", [])
+                    cards = entity.get("dashcards", [])
```
gouline, line 129:
```suggestion
                    cards = entity.get("dashcards", entity.get("ordered_cards", []))
```

This appears to be a change in [Metabase 48](https://www.metabase.com/releases/metabase-48) in January, so a lot of people may still be running older versions

gouline (changes requested):

gouline (approved):

PR #251, opened 2024-04-09
Add default descriptions to model export

jurasan:


gouline:
Thanks for your contribution, but I don't see value in this functionality. Additional layer of lookups between the dbt project and exporter goes against the project's mission of simply mapping dbt features to their Metabase counterparts. Your implementation also appears narrowly focused on your specific setup and lacks CLI support. As an alternative, consider preprocessing your dbt project with any defaults before feeding it into dbt-metabase.

In future, please raise an issue for any non-trivial changes documenting the use case and proposed implementation first. Once it has been discussed, make sure any pull requests you open are well documented.

jurasan:
:) Sorry, it was actually meant as a PR into our fork. I just fat-fingered

Issue #255, opened 2024-04-22
Auto-Create Metabase Questions from dbt Models


korbash:
# Auto-Create Metabase Questions from dbt Models

Creating SQL queries with multiple filters can be challenging. One of the most useful types of filters in Metabase is the [Field Filters](https://www.metabase.com/learn/sql-questions/field-filters). Interestingly, it's possible to integrate these filters directly from dbt models!

We can enhance the `schema.yml` file by adding filter features to columns intended to be used as filters. Here's an example of how to define these attributes in the metadata section:

```yaml
models:
  - name: orders
    columns:
      - name: order_date
        meta:
          metabase.filter:
            base-type: "type/Date"
            widget-type: "date/range"
```
To utilize these filters in the dbt models that are to be transformed into Metabase queries, reference the column in your model as follows:

```sql
WHERE '__filter__.<model_name>.<column_name>' = '__filter__.<model_name>.<column_name>'
```
After this setup, the dbt-metabase tool can automatically generate Metabase questions using the information from the `manifest.json` and the compiled models code.

gouline:
Current focus of the tool is syncing metadata (like tables and exposures) between dbt and Metabase. Generating questions goes well beyond that into analytics.

gouline:
Tag me on the pull request in your fork when you do, I'd be curious how you implement it.

PR #256, opened 2024-05-02
Metabase cards with filters

korbash:
Very first implementation of Metabase cards with filters from dbt models.
Idea discussed in https://github.com/gouline/dbt-metabase/issues/255

@gouline, you asked me to mention you.


gouline:
Thanks, I understand your idea better now. This confirms my suspicion that it goes beyond the scope of the project. Additionally, this early implementation would require a lot of work before it can be considered for contribution, so it might be easier to keep it on your fork.

PR #267, opened 2024-07-11
Skip tables that are technical not just hidden

chrismdp:
Tables are hidden with any non-nil value - we hide our tables with  `technical` so would be useful to update this.

### DIFF
--- a/dbtmetabase/_models.py
+++ b/dbtmetabase/_models.py
@@ -100,7 +100,7 @@ class ModelsMixin(metaclass=ABCMeta):
 
                     field = table.get("fields", {}).get(column_name)
                     if not field:
-                        if table.get("visibility_type") == "hidden":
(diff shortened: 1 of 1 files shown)


dbtmetabase/_models.py
```diff
@@ -100,7 +100,7 @@ def export_models(
 
                     field = table.get("fields", {}).get(column_name)
                     if not field:
-                        if table.get("visibility_type") == "hidden":
+                        if table.get("visibility_type") is not None:
```
gouline, line 103: Not the only instance of "hidden" comparison you need to fix

gouline (changes requested):

dbtmetabase/_models.py
chrismdp: ah sorry - been a busy week - I can see you've merged this anyway so thanks!

PR #273, opened 2024-08-07
Add support for Metabase instances deployed on GCP protected by Identity-Aware Proxy (IAP)

jefersonmsantos:
- Add support to pass GCP service account to argument `--gcp-iap-service-account` to perform authentication when Metabase instance is deployed on GCP and protected by Identity-Aware Proxy (IAP).
  - service account must have `IAP-secured Web App User` role on Metabase Application
  - it will generate a short living jwt token for each Metabase API endpoint to be called
- API-KEY from metabase is still necessary on this scenario.


gouline:
Sorry, this implementation is too specific to infrastructure outside of Metabase, I'm not prepared to maintain it and require users to install additional dependencies for your isolated usage (imagine others contributing AWS, Azure, etc).

You are welcome to use `http_headers` parameter to pass `Authorization: Bearer` token.

jefersonmsantos:
> Sorry, this implementation is too specific to infrastructure outside of Metabase, I'm not prepared to maintain it and require users to install additional dependencies for your isolated usage (imagine others contributing AWS, Azure, etc).
> 
> You are welcome to use `http_headers` parameter to pass `Authorization: Bearer` token.

Problem is that map requires a token generated for each endpoint, not for subdomains. So we can't use http_headers.

Issue #280, opened 2024-10-09
Additional metadata to understand "usability" of a card


codingcyclist:
Hi there,

I'm a big fan of `dbt-metabase`! Especially the `exposures` feature has been a complete game changer to avoid deploying model changes that would break metabase cards.

With a growing number of cards in Metabase, my team figured that it's not only important to understand by which cards our DBT models are referenced , but also whether these cards are still being used. Hence, I was thinking to add the `last_used_at` and `average_query_time` fields from the metabase card API to the exposure metadata. They strike me as a good proxy for whether a card is still usable (query time) and whether it gets used (last used timestamp)

I've already done this on a local fork (see screenshot) of the repo and would be happy to contribute it. Wdyt @gouline?

<img width="1029" alt="Screenshot 2024-10-09 at 17 49 49" src="https://github.com/user-attachments/assets/c53a5d59-4e69-4ca5-a324-5da5521bb9e8">

PR #281, opened 2024-10-09
Additional metadata to understand "usability" of a card

codingcyclist:
See https://github.com/gouline/dbt-metabase/issues/280

### DIFF
--- a/dbtmetabase/_exposures.py
+++ b/dbtmetabase/_exposures.py
@@ -112,7 +112,9 @@ class ExposuresMixin(metaclass=ABCMeta):
                 depends = set()
                 native_query = ""
                 header = ""
-
+                average_query_time_ms = None
+                last_used_at_utc = None
+                
                 entity: Mapping
                 if item["model"] == "card":
                     card_entity = self.metabase.find_card(uid=item["id"])
@@ -128,6 +130,8 @@ class ExposuresMixin(metaclass=ABCMeta):
                     result = self.__extract_card_exposures(ctx, card=entity)
                     depends.update(result["depends"])
                     native_query = result["native_query"]
+                    average_query_time_ms = entity.get('average_query_time')
+                    last_used_at_utc = entity['last_used_at']
 
                 elif item["model"] == "dashboard":
                     dashboard_entity = self.metabase.find_dashboard(uid=item["id"])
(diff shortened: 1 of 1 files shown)


dbtmetabase/_exposures.py
```diff
@@ -128,6 +130,8 @@ def extract_exposures(
                     result = self.__extract_card_exposures(ctx, card=entity)
                     depends.update(result["depends"])
                     native_query = result["native_query"]
+                    average_query_time_ms = entity.get('average_query_time')
+                    last_used_at_utc = entity['last_used_at']
```
gouline, line 134: Should be an optional `get()` as well

dbtmetabase/_exposures.py
```diff
@@ -112,7 +112,9 @@ def extract_exposures(
                 depends = set()
                 native_query = ""
                 header = ""
-
+                average_query_time_ms = None
+                last_used_at_utc = None
```
gouline, line 116: Since you're retrieving them as-is from Metabase, keep the naming identical (i.e. no `_ms` and `_at_utc`)

dbtmetabase/_exposures.py
```diff
@@ -346,6 +354,11 @@ def __format_exposure(
             + f"Created On: __{created_at}__"
         )
 
+        if average_query_time_ms:
+            average_query_time_s = average_query_time_ms // 1000
+            average_query_time = f"{int(average_query_time_s // 60)}:{str(int(average_query_time_s % 60)).zfill(2)}"
+        else:
+            average_query_time = "n.a."
```
gouline, line 361: If not provided, shouldn't be included at all

dbtmetabase/_exposures.py
```diff
@@ -360,6 +373,10 @@ def __format_exposure(
                 "email": creator_email,
             },
             "depends_on": list(depends_on),
+            "meta": {
+                "avg_query_time": average_query_time,
+                "last_used_at_utc": last_used_at_utc or 'n.a.'
```
gouline, line 378: Same here, if not provided, shouldn't be included at all (build up `meta` conditionally as a dictionary variable and only set it when not empty)

dbtmetabase/_exposures.py
```diff
@@ -346,6 +354,11 @@ def __format_exposure(
             + f"Created On: __{created_at}__"
         )
 
+        if average_query_time_ms:
+            average_query_time_s = average_query_time_ms // 1000
+            average_query_time = f"{int(average_query_time_s // 60)}:{str(int(average_query_time_s % 60)).zfill(2)}"
```
gouline, line 359: If the original value is in milliseconds (i.e. 1000th), why are you zfill'ing with only 2 and not 3?

gouline (changes requested):
You haven't run the linter/formatter, please do this before pushing any commits

PR #284, opened 2024-10-12
feat: query average time and last usage metadata

### LINKED ISSUE #280: Additional metadata to understand "usability" of a card

gouline:
- Card `average_query_time` (in milliseconds) presented as `0:00.000` in exposure metadata
- Card `last_used_at` (ISO 8601 timestamp) presented as-is in exposure metadata
- Update exposure fixtures to reflect these
- Continues #281
- Resolves #280

### DIFF
--- a/dbtmetabase/_exposures.py
+++ b/dbtmetabase/_exposures.py
@@ -112,7 +112,9 @@ class ExposuresMixin(metaclass=ABCMeta):
                 depends = set()
                 native_query = ""
                 header = ""
-
+                average_query_time_ms = None
+                last_used_at_utc = None
+                
                 entity: Mapping
                 if item["model"] == "card":
                     card_entity = self.metabase.find_card(uid=item["id"])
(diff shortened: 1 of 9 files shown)


codingcyclist:
Thanks for fixing the comments <3 @gouline! I was just about to get to it

PR #291, opened 2024-11-19
feat: allow column description to be overwritten

sabino:
This allows `description` to be set using `metabase.description` at column level, prioritizing it when set. 

Feel free to suggest any possible changes, but just to give you an idea of our use case. We might (and need) different descriptions for columns due to: language, size (dbt documentation allows for more flexibility).

I might update the PR to include the same behavior at Table level.

### DIFF
--- a/dbtmetabase/manifest.py
+++ b/dbtmetabase/manifest.py
@@ -26,6 +26,7 @@ _META_NS = "metabase"
 _COMMON_META_FIELDS = [
     "display_name",
     "visibility_type",
+    "description",
 ]
 # Must be covered by Column attributes
 _COLUMN_META_FIELDS = _COMMON_META_FIELDS + [
@@ -136,14 +137,17 @@ class Manifest:
         schema: str,
         relationship: Optional[Mapping],
     ) -> Column:
-        column = Column(
-            name=manifest_column.get("name", ""),
-            description=manifest_column.get("description"),
-            **self._scan_fields(
-                manifest_column.get("meta", {}),
+        
+        scanned_fields = self._scan_fields(
+            manifest_column.get("meta", {}),
                 fields=_COLUMN_META_FIELDS,
                 ns=_META_NS,
-            ),
+        )
+        
+        column = Column(
+            name=manifest_column.get("name", ""),
+            description=scanned_fields.get("description", manifest_column.get("description")),
+            **{key: value for key, value in scanned_fields.items() if key != "description"},
         )
 
         self._set_column_relationship(


gouline:
Yeah, if you're adding it for columns, it should be there for tables too. Add some unit tests as well.

tests/fixtures/manifest-v12.json
```diff
@@ -36,7 +36,9 @@
                 "schema": null,
                 "database": null,
                 "tags": [],
-                "meta": {},
+                "meta": {
```
gouline, line 39: This file gets overwritten from the sandbox dbt project, easiest thing would be to put these in one of the older version manifests that don't get overwritten

gouline (changes requested):
CI is failing, `make check test` should succeed locally before pushing any commits

tests/fixtures/manifest-v12.json
sabino: Can I just change the `sandbox` content definition to match this file then? And update older versions as well, so that they all match

gouline: You could, but that might involve debugging other tests if the manifest has other updates, it would be easier if you just updated one older version of the manifest (and corresponding assertions). They don't all need to match, one is sufficient to test this functionality somewhere

dbtmetabase/manifest.py
```diff
@@ -110,6 +111,12 @@ def _read_model(
             for column in manifest_model.get("columns", {}).values()
         ]
 
+        scanned_fields = self._scan_fields(
+            manifest_model.get("meta", {}),
+            fields=_MODEL_META_FIELDS,
+            ns=_META_NS,
+        )
+
```
gouline, line 119:
```suggestion
        description = meta.pop("description", manifest_model.get("description"))
```

dbtmetabase/manifest.py
```diff
@@ -110,6 +111,12 @@ def _read_model(
             for column in manifest_model.get("columns", {}).values()
         ]
 
+        scanned_fields = self._scan_fields(
```
gouline, line 114:
```suggestion
        meta = self._scan_fields(
```

dbtmetabase/manifest.py
```diff
@@ -118,16 +125,18 @@ def _read_model(
             alias=manifest_model.get(
                 "alias", manifest_model.get("identifier", manifest_model["name"])
             ),
-            description=manifest_model.get("description"),
+            description=scanned_fields.get(
+                "description", manifest_model.get("description")
+            ),
```
gouline, line 130:
```suggestion
            description=description,
```

dbtmetabase/manifest.py
```diff
@@ -118,16 +125,18 @@ def _read_model(
...
-            **self._scan_fields(
-                manifest_model.get("meta", {}),
-                fields=_MODEL_META_FIELDS,
-                ns=_META_NS,
-            ),
+            **{
+                key: value
+                for key, value in scanned_fields.items()
+                if key != "description"
+            },
```
gouline, line 139:
```suggestion
            **meta,
```

dbtmetabase/manifest.py
```diff
@@ -136,14 +145,22 @@ def _read_column(
         schema: str,
         relationship: Optional[Mapping],
     ) -> Column:
+        scanned_fields = self._scan_fields(
```
gouline, line 148:
```suggestion
        meta = self._scan_fields(
```

dbtmetabase/manifest.py
```diff
@@ -136,14 +145,22 @@ def _read_column(
...
+            manifest_column.get("meta", {}),
+            fields=_COLUMN_META_FIELDS,
+            ns=_META_NS,
+        )
+
```
gouline, line 153:
```suggestion
        description = meta.get("description", manifest_column.get("description"))
```

dbtmetabase/manifest.py
```diff
@@ -136,14 +145,22 @@ def _read_column(
...
         column = Column(
             name=manifest_column.get("name", ""),
-            description=manifest_column.get("description"),
-            **self._scan_fields(
-                manifest_column.get("meta", {}),
-                fields=_COLUMN_META_FIELDS,
-                ns=_META_NS,
+            description=scanned_fields.get(
+                "description", manifest_column.get("description")
             ),
```
gouline, line 158:
```suggestion
            description=description,
```

dbtmetabase/manifest.py
```diff
@@ -136,14 +145,22 @@ def _read_column(
...
+            **{
+                key: value
+                for key, value in scanned_fields.items()
+                if key != "description"
+            },
```
gouline, line 163:
```suggestion
            **meta,
```

gouline (approved):

PR #316, opened 2025-04-02
Only update the settings attribute on fields if there are setting changes

### LINKED ISSUE #315: Column settings will update every run if any are set in Metabase
When the API returns a number_style/decimals/etc. for a column, every run will write them back again.

willbryant:
The code already tried to do this, but it actually checked if there were any entries in the column settings attribute - including those received from the API itself.

closes #315


gouline (approved):
Thanks. You don't need to create an issue if you're fixing it yourself in a pull request.

Issue #331, opened 2025-09-18
Support for a Databricks connection with multiple catalogs


AlizaminJ:
First of all, I am not quite sure if it is the issue. Hoping for understanding as there is no meaningful log in Metabase admin to supplement the issue here.

We have recently enabled a relatively new feature in Metabase, multiple catalogs within a single Databricks connection. As far as I see it is not supported, as dbt docs are not propagated to Metabase.  I tried to adjust with `database/schema/model_filter`, but no result. It either silently fails, or complains with something like "WARNING:dbtmetabase._models:Table 'BRONZE.MY_MODEL' not in schema 'BRONZE'

Maybe worth mentioning, the same table appears differently depending on the connection type. Upon search, "MY_MODEL" in Metabase console:
MY_MODEL - CONNECTION (CATALOG.SCHEMA) --> connection with multiple catalog
MY_MODEL - CATALOG(SCHEMA) --> connection with single catalog

PR #333, opened 2025-10-03
Add multi-database support

AlizaminJ:
Ref to [issue](https://github.com/gouline/dbt-metabase/issues/331).

This PR adds Databricks multi-catalog support to dbt-metabase with full backward compatibility. Uses PostgreSQL-first approach with Databricks fallback.

### DIFF
--- a/dbtmetabase/_models.py
+++ b/dbtmetabase/_models.py
@@ -390,8 +390,7 @@ class ModelsMixin(metaclass=ABCMeta):
                     fk_target_field_id = fk_target_field.get("id")
                     if fk_target_field.get(semantic_type_key) != "type/PK":
                         _logger.info(
-                            "Field '%s' will be updated as primary key "
-                            "for foreign key '%s'",
+                            "Field '%s' will be updated as primary key for foreign key '%s'",
                             fk_target_field_label,
                             column_label,
                         )


gouline (changes requested):
- Do not add Databricks to sandbox. It's a lean local environment for end-to-end testing and no databases other than Postgres, especially those requiring accounts and SDKs (e.g. Databricks, Snowflake), should be in it.
- This is clearly vibe coded, which is fine, but revert all irrelevant changes before you push. From a brief look, only _models.py and test_database_schema_table_format.py are relevant to your feature.

gouline:
Unit tests and run it on your own dbt project that I presume uses multi-catalog Databricks to verify everything works.

AlizaminJ:
> Unit tests and run it on your own dbt project that I presume uses multi-catalog Databricks to verify everything works.

Unit tests are fine ✅ .  What do you mean with "run it on your own dbt project"? I beg your pardon, this is not about dbt project with databricks but database connection type in metabase that made it stopped working, see the flag ***Enable multiple catalogs***. Not quite sure how I could export my models outside this project? I bet it will be quite challenging to debug stuff.

Having said that as you already mentioned _models.py and test_database_schema_table_format.py  are what I need to make it work for now ✅ . I am just curious how your setup would otherwise look like in this case, something that I can use next time.
 
image

gouline:
Regardless of whether it impacts dbt or Metabase, my expectation is that you run dbt-metabase on your branch against your actual dbt project, Metabase instance and Databricks account to verify your implementation works as intended, without relying on sandbox.

dbtmetabase/_models.py
```diff
@@ -171,11 +176,22 @@ def _export_model(
...
+        # Fallback for multi-catalog connections: try catalog.schema.table format
+        if not api_table and model.database:
+            database_schema_table_key = model.alias_path.upper()
+            api_table = ctx.tables.get(database_schema_table_key)
+            if api_table:
+                _logger.debug(
+                    "Using multi-catalog format for table: %s",
+                    database_schema_table_key,
+                )
+                table_key = database_schema_table_key
```
gouline: Same comments as above

dbtmetabase/_models.py
```diff
@@ -339,17 +352,43 @@ def __export_column(
...
                     field_key=fk_target_field_name,
                 )
+
+                # Fallback for multi-catalog connections: try catalog.schema.table format
+                if (
+                    not fk_target_field
+                    and "." in table_key
+                    and fk_target_table_name.count(".") < 2
+                ):
+                    catalog_part = table_key.split(".")[0]
```
gouline: As above, database, not catalog part

dbtmetabase/_models.py
```diff
@@ -461,6 +505,24 @@ def _get_metabase_tables(self, database_id: str) -> Mapping[str, MutableMapping]
...
+
+            # Extract database name from table["db"]
+            if table.get("db"):
+                db_parts = str(table["db"]).split(".")
+                if len(db_parts) >= 2:
+                    # Multi-database connections: "database.schema" format
+                    database_name = db_parts[0].upper()
+                else:
+                    # Single database: just the database name
+                    database_name = str(table["db"]).upper()
```
gouline:
```suggestion
            if table_db := table.get("db"):
                db_parts = str(table_db).split(".")
                if len(db_parts) >= 2:
                    # Multi-database connections: "database.schema" format
                    database_name = db_parts[0].upper()
                else:
                    # Single database: just the database name
                    database_name = str(table_db).upper()
```

gouline (approved):

PR #337, opened 2025-12-04
Support exposure extraction for MBQL 5 cards

reidab:
As of Metabase 0.57, cards are serialized in MBQL 5 format:

> MBQL queries (in Cards and elsewhere) are now serialized as MBQL 5 as opposed to MBQL 4 (aka legacy MBQL) in the application database and in REST API responses. While we do not officially support editing or introspection of MBQL via the REST API (please treat it as an opaque object), to support existing usages the GET /api/card/:id endpoint can return the Card dataset_query as MBQL 4 if you include the query parameter ?legacy-mbql=true.
-- https://www.metabase.com/docs/latest/developers-guide/api-changelog#metabase-0570

This PR updates card parsing to support the new format.

Now, It wasn't until I was submitting this PR that I noticed that last part about being able to request legacy mbql with `?legacy-mbql=true`. To support Metabase 0.57+, you either need the changes in this PR or to switch to using that query param when fetching cards with the existing legacy implementation.

### DIFF
--- a/dbtmetabase/_exposures.py
+++ b/dbtmetabase/_exposures.py
@@ -251,6 +251,14 @@ class ExposuresMixin(metaclass=ABCMeta):
         """Extracts exposures from Metabase questions."""
 
         dataset_query = card.get("dataset_query", {})
+
+        # Check for newer MBQL5 format
+        lib_type = dataset_query.get("lib/type")
+        if lib_type == "mbql/query":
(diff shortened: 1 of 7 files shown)


reidab:
Yep, this maintains the existing parsing and adds support for the new format. The SQL parsing for native queries is shared between both paths.

gouline (approved):

PR #362, opened 2026-03-27
Fix for #360: Handle 403 errors for card and user access

### LINKED ISSUE #360: Handle 403 responses in find_card and find_user to prevent exposure generation from failing
For cards in personal collections used in dashboards, we hit a 403 error, and so the exposures file was not created. 

Adding a 403 guard (similarly to how the 404 is already handled) to metabase.py logs a warning and continues to exposure generation:

`find_card`

```
if error.response.status_code == 404:
    _logger.warning("Card '%s' not found", uid)
    return None
# add:
if error.response.status_code == 403:
    _logger.warning("Card '%s' not accessible", uid)
    return None
raise
```

`find_user`

```
if error.response.status_code == 404:
    _logger.warning("User '%s' not found", uid)
    return None
# add:
elif error.response.status_code == 403:
    _logger.warning("No permission to fetch user '%s', skipping creator info", uid)
    return None
elif ...:
```

natmorawska:
- Handle 403 responses in `find_card` and `find_user` to prevent exposure generation from failing when cards are in personal collections.
- Resolves #360

### DIFF
--- a/dbtmetabase/metabase.py
+++ b/dbtmetabase/metabase.py
@@ -199,7 +199,7 @@ class Metabase:
                 _logger.warning("User '%s' not found", uid)
                 return None
             elif error.response.status_code == 403:
-                _logger.warning("No permission to fetch user '%s', skipping creator info", uid)
+                _logger.warning("User '%s' not accessible", uid)
                 return None
             elif (
                 error.response.status_code == 400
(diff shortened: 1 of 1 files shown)


dbtmetabase/metabase.py
```diff
@@ -195,6 +198,9 @@ def find_user(self, uid: str) -> Mapping | None:
             if error.response.status_code == 404:
                 _logger.warning("User '%s' not found", uid)
                 return None
+            elif error.response.status_code == 403:
+                _logger.warning("No permission to fetch user '%s', skipping creator info", uid)
```
gouline:
```suggestion
                _logger.warning("User '%s' not accessible", uid)
```

gouline (approved):

PR #393, opened 2026-07-14
Remap foreign key display values to a target's declared display field

joakimwar:
## Motivation

`dbt-metabase models` sets `fk_target_field_id`, which draws the join relationship in Metabase's data model. But in tables and dashboards Metabase still shows the **raw foreign key id**. Which column is shown in its place is controlled by a separate Metabase resource — a field *dimension* (`POST /api/field/:id/dimension`) — that `schema.yml` had no way to express. Until now that gap had to be filled with a standalone post-sync script.

## What this adds

A `metabase.display_field` column meta marker. Set it once, on a model's human-readable column:

```yaml
- name: campaign_name
  config:
    meta:
      metabase.display_field: true
```

Every foreign key that resolves to that table — however the relationship was defined (constraints, relationship tests, or `fk_target_table`/`fk_target_field` meta) — has its dimension remapped to show `campaign_name` instead of the raw id. No per-key annotation is needed.

### Behavior details

- **Read from the full manifest**, not the filtered set: a foreign key can point at a table that a `--schema`/`--model` filter excluded from export, but its declared display field is still needed to remap the columns pointing at it.
- **Idempotent**: a dimension already pointing at the target field is left untouched.
- **One per table**: if several columns are marked, the first is used (with a warning).
- Reuses the existing batched-update mechanism via a new `field_dimension` update kind and `Metabase.update_field_dimension`.

## Changes

- `manifest.py`: recognize `metabase.display_field` (`Column.display_field`).
- `metabase.py`: `update_field_dimension()` → `POST /api/field/:id/dimension`.
- `_models.py`: build the display-field map, remap FK dimensions during column export, dispatch `field_dimension` updates.
- `README.md`: new "Display Values" section.
- `tests/test_models.py`: cover the map build and the apply/skip paths.

🤖 Generated with [Claude Code](https://claude.com/claude-code)

### DIFF
--- a/README.md
+++ b/README.md
@@ -116,6 +116,21 @@ Native [relationship tests](https://docs.getdbt.com/reference/resource-propertie
 
 You can provide `fk_target_table` as `schema_name.table_name` or just `table_name` to use the current schema. If your model has an alias, provide that alias rather than the original name.
 
+### Display Values
+
+By default a foreign key column shows the raw id it points at. Metabase can instead display a human-readable column from the target table (its "display value"). Mark that column once, on the target model, with `metabase.display_field`:
+
+```yaml
+- name: campaign_name
+  config:
+    meta:
+      metabase.display_field: true
+```
+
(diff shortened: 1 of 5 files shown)


gouline:
Not worth it for the amount of changes.

